PR #28, opened 2026-02-10
fix(sync): backfill derived fields and advance filtered cursors

kunickiaj:
## Summary
- backfill tags_text and vectors for memory items touched by incoming sync upserts
- allow targeted backfill by memory id to avoid broad backfill runs on every sync
- advance last_applied_cursor when peer returns ops=[] with skipped items and a forward next_cursor

## Validation
- uv run pytest tests/test_sync_daemon.py tests/test_sync_protocol.py tests/test_store.py -k "backfill_tags_text or backfill_vectors or sync_once"

Refs: opencode-mem-hu0, opencode-mem-ba2

### DIFF
--- a/opencode_mem/cli_app.py
+++ b/opencode_mem/cli_app.py
@@ -760,26 +760,35 @@ def sync_status(db_path: str = typer.Option(None, help="Path to SQLite database"
 
 @sync_app.command("pair")
 def sync_pair(
-    accept: str | None = typer.Option(None, help="Accept pairing payload (JSON)"),
+    accept: str | None = typer.Option(None, help="Accept pairing payload JSON from another device"),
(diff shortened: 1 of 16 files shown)


kunickiaj:
Addressing review note on empty accept payload: this is now fixed in base branch PR #27 (commit 3c535a9).  now fails fast when / is explicitly provided with empty payload (file or stdin), with regression tests for both cases. I rebased this PR onto the updated base so the fix is included transitively.

kunickiaj:
Follow-up clarification: the fix covers explicit accept mode using flags like accept and accept-file with empty input. In that case sync pair now exits non-zero instead of falling back to payload generation. Tests were added for both empty file and empty stdin.

PR #30, opened 2026-02-10
fix(viewer): harden origin checks and request limits

kunickiaj:
## Summary
- replace prefix-based Origin checks with strict parsed loopback origin validation
- require explicit Origin on mutating sync POST/DELETE routes while preserving safe non-browser raw-events ingestion
- add request-size limit for raw-events and validate sync attempts limit parsing

## Validation
- uv run pytest tests/test_viewer_http.py tests/test_sync_viewer_api.py tests/test_viewer_routes_raw_events.py tests/test_store.py::test_viewer_accepts_raw_events

Refs: opencode-mem-a7n, opencode-mem-3rt, opencode-mem-5ej, opencode-mem-juh

### DIFF
--- a/opencode_mem/viewer.py
+++ b/opencode_mem/viewer.py
@@ -15,6 +15,7 @@ from .observer import _load_opencode_config
 from .raw_event_flush import flush_raw_events  # noqa: F401
 from .store import MemoryStore
 from .viewer_http import (
+    MissingOriginPolicy,
     read_json_body,
     reject_cross_origin,
     send_bytes_response,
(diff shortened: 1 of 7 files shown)


kunickiaj:
Thanks for the catch — fixed. Malformed Origin values (including invalid IPv6 forms like ) are now treated as untrusted and return a clean 403 instead of raising during parsing. Added regression coverage in tests/test_viewer_http.py.

kunickiaj:
Fixed: malformed Origin values now fail closed and return 403 instead of raising parser errors. This includes invalid IPv6 forms and malformed netloc values. Added regression coverage in tests/test_viewer_http.py.

PR #699, opened 2026-04-10
fix(core): auto-bootstrap fresh databases in MemoryStore constructor

kunickiaj:
## Summary

Regression follow-up to #575. That PR taught \`codemem db init\` and
\`codemem serve start\` to auto-bootstrap a fresh SQLite file, but
didn't hoist the fix into the shared \`MemoryStore\` constructor or
the \`directEnqueue\` hook path. So every other entry point —
\`mcp-server\`, \`claude-hook-ingest\`, the CLI commands
(\`stats\`, \`search\`, \`recent\`, \`memory\`, \`embed\`, etc.) —
still called \`assertSchemaReady\` (or raw \`db.prepare("SELECT ... FROM raw_events")\`)
and threw on an uninitialized DB.

**User-visible symptom:** installing the plugin into a fresh
environment — notably a Claude Desktop / Cowork sandbox VM whose
\`~/.codemem\` is wiped on every session — produced an empty
\`mem.sqlite\` via \`connect()\` but then the very first MCP tool
call or hook invocation crashed because no schema existed. The user
had to run \`codemem db init\` manually before anything worked.

## Fix

Introduces a small \`ensureSchemaBootstrapped(db)\` helper in
\`schema-bootstrap.ts\`:

\`\`\`ts
export function ensureSchemaBootstrapped(db: Database): void {
    if (getSchemaVersion(db) === 0) {
        bootstrapSchema(db);
    }
}
\`\`\`

Exported from \`@codemem/core\`. Called from two places:

1. **\`MemoryStore\` constructor** (\`packages/core/src/store.ts\`) —
   runs right after \`loadSqliteVec\` and before
   \`assertSchemaReady\`. Matches the \`initDatabase()\` gate pattern
   exactly: a fresh file gets the full schema DDL, an
   already-initialized file is left alone. Idempotent with the
   existing \`db init\` / \`serve start\` paths.

2. **\`directEnqueue\`** (\`packages/cli/src/commands/claude-hook-ingest.ts\`) —
   runs right after \`loadSqliteVec\` and before the raw
   \`SELECT 1 FROM raw_events ...\` prepare. This closes the
   MCP-server-startup vs. hook-spawn race: the hook script is a
   separate CLI process and we can't rely on the MCP server having
   bootstrapped the DB first.

Every other entry point that touches the DB routes through
\`MemoryStore\` (mcp-server, claude-hook-inject's
\`buildLocalPack\`, \`flushBoundaryRawEvents\`, the read-only CLI
[...]

PR #700, opened 2026-04-10
chore(plugin): surface hook stderr and pin codemem version in scripts

kunickiaj:
## Summary

Two small hook-script improvements driven by the Cowork sandbox
investigation where \`PostToolUse\` hooks were consistently logging
the generic line \`codemem claude-hook-ingest failed via npx\` with
zero diagnostic context, making root-cause analysis impossible.

Stacked on #699 (auto-bootstrap fix) so 0.25.3 lands both together.

## ingest-hook.sh

Replaces the two inline \`command ... >/dev/null 2>&1\` attempts
with a shared \`run_ingest\` helper that:

- Captures stderr to a tempfile (with a graceful fallback if mktemp
  is unavailable, logging an explicit \`<unavailable: mktemp failed>\`
  marker so the degraded path is visible).
- Measures wall-clock time in milliseconds via a portable \`now_ms\`
  helper: \`EPOCHREALTIME\` on bash 5+ → GNU \`date +%s%N\` when
  available → \`date +%s * 1000\` second-precision fallback. No
  literal-\`N\` corruption on BSD \`date\` (macOS).
- On failure, logs a structured line:
  \`codemem claude-hook-ingest failed via <label> rc=<N> ms=<N> stderr=<first 400 chars>\`
  The \`stderr=\` field is always last because the excerpt is raw
  freeform text; future \`key=value\` fields must be inserted before it.
- Strips ALL C0 control characters + DEL from the excerpt via
  \`LC_ALL=C tr '\\000-\\037\\177' ' '\`. Kills ANSI log-forging,
  null-byte log truncation, and weird Unicode nonsense in a single
  cheap pipe.
- On success, stays silent (cleans up the tempfile).

Adds a \`resolve_pinned_version\` helper that pins the
\`npx -y codemem\` invocation to the version declared in
\`plugin.json\`. Uses \`jq\` when available; falls back to a sed
pattern so Cowork sandbox VMs (which don't ship jq by default) still
get version pinning instead of silently degrading to
\`npx -y codemem@latest\`. The Cowork diagnosis found a plugin
declaring \`0.25.2\` but npx silently running \`0.22.4\` out of an
older cache — this prevents that drift.

## user-prompt-hook.sh

Drops the redundant \`|| true\` guards around the async ingest
invocation. \`ingest-hook.sh\` now handles all of its own error
logging via the new \`run_ingest\` path, so the outer wrapper doesn't
need to swallow exit codes — nohup backgrounds the child regardless
of its return value.

## Smoke tests (local, macOS bash 3.2)

Failure path with stub codemem emitting ANSI + null + newlines:

\`\`\`
$ stub stderr: "Error: db locked \\x1b[31mRED\\x1b[0m\\nsecond line\\tafter\\x00null=inside\\n"
2026-04-10T22:20:22Z codemem claude-hook-ingest failed via codemem binary rc=42 ms=297 stderr=Error: db locked  [31mRED [0m second line after tab null byte=inside
\`\`\`

Note: ESC (0x1B) is stripped leaving benign text \`[31m\`; null byte
stripped (no log truncation); newline/tab flattened to spaces;
\`ms=297\` is a real measured value.

Success path with real \`codemem claude-hook-ingest\`: \`exit=0\`,
no new log lines from this script.

Version pin resolution (with and without \`jq\` on PATH) against
\`plugins/claude/.claude-plugin/plugin.json\`: both resolve to
\`0.25.2\`.

Now-ms portability (simulated BSD \`date\`): falls through correctly
to seconds × 1000, no \`N\`-char arithmetic corruption.

## Review feedback addressed

Gordon Ramsay reviewed the initial version and flagged four issues,
all fixed here:

1. **\`date +%s%N\` broken on BSD/macOS** → \`now_ms\` helper with
   EPOCHREALTIME / GNU date / seconds-fallback chain.
2. **Stderr excerpt strips only whitespace, not ANSI / null / control
   chars** → \`LC_ALL=C tr '\\000-\\037\\177' ' '\` strips all C0 + DEL.
3. **jq dependency silently falls back to \`latest\`, defeating the
   pin** → sed pattern fallback when jq is missing.
4. **mktemp failure path silently uses old broken behavior** → now
   logs an explicit \`<unavailable: mktemp failed>\` marker.

Nice-to-haves also applied: locals initialized with default values
for \`set -u\` defensiveness, explicit comment that \`stderr=\` must
stay the last field.

## Type of Change

- [x] 🔧 Maintenance / observability (no behavior change on happy path)

## Testing

- [x] \`bash -n\` on both scripts — clean
- [x] Failure path smoke test with stubbed codemem (control-char
  sanitization, ms timing, rc capture, exit code propagation)
- [x] Success path smoke test with real codemem (silent, exit 0)
- [x] Version pin resolution with and without jq
- [x] \`now_ms\` fallback verified against simulated BSD \`date\`
- [x] \`pnpm run test\` — 1193 passed (no script tests, but confirms
  nothing in the workspace depends on the old hook behavior)

## Checklist

- [x] No new crashes on any error path
- [x] \`stderr=\` stays last field so freeform excerpt doesn't break
  parsers
- [x] Control-char sanitization prevents log forging
- [x] jq dependency is optional, not required
- [x] No private refs / internal hostnames in code or commit message
- [x] Stacked on #699 (auto-bootstrap fix)

### DIFF
--- a/plugins/claude/scripts/ingest-hook.sh
+++ b/plugins/claude/scripts/ingest-hook.sh
@@ -16,6 +16,55 @@ log_line() {
 SCRIPT_DIR="$(CDPATH= cd -- "$(dirname -- "$0")" && pwd)"
 PLUGIN_ROOT="${CLAUDE_PLUGIN_ROOT:-$(CDPATH= cd -- "${SCRIPT_DIR}/.." && pwd)}"
 PLUGIN_MANIFEST="${PLUGIN_ROOT}/.claude-plugin/plugin.json"
+
+# Portable millisecond-resolution clock.
+#
+# - bash 5+ exposes `EPOCHREALTIME` (fractional epoch seconds). Zero subshells.
+# - GNU date supports `%N` for nanoseconds; BSD date (macOS default) does not
+#   and would print a literal `N`, breaking downstream arithmetic.
+# - Fall back to second-precision epoch (ms ends in 000) if neither is
+#   available.
+now_ms() {
+  if [ -n "${EPOCHREALTIME:-}" ]; then
+    # "1744300000.123456" → "1744300000123456" → drop last 3 digits = ms
+    local raw="${EPOCHREALTIME/./}"
+    printf '%s' "${raw%???}"
+    return
+  fi
+  # Probe once per call — cheap vs forking date twice in the caller.
+  local probe
+  probe="$(date +%N 2>/dev/null)"
+  if [ "${probe}" != "N" ] && [ -n "${probe}" ] && [ "${probe}" != "%N" ]; then
+    printf '%s' "$(( $(date +%s%N) / 1000000 ))"
+    return
+  fi
+  printf '%s' "$(( $(date +%s) * 1000 ))"
+}
+
+# Resolve the codemem version to pin `npx -y codemem@<version>` against.
+# Reads `plugins/claude/.claude-plugin/plugin.json`; intentionally uses a
+# sed fallback instead of jq so the pin works in minimal environments (Cowork
+# sandbox VMs don't ship jq by default). The plugin manifest is controlled by
+# this repo, so the narrow sed pattern is sufficient.
+resolve_pinned_version() {
+  if [ ! -r "${PLUGIN_MANIFEST}" ]; then
+    printf 'latest'
+    return
+  fi
+  local v=""
+  if command -v jq >/dev/null 2>&1; then
+    v="$(jq -r '.version // empty' "${PLUGIN_MANIFEST}" 2>/dev/null)"
+  fi
+  if [ -z "${v}" ] || [ "${v}" = "null" ]; then
+    v="$(sed -n 's/.*"version"[[:space:]]*:[[:space:]]*"\([^"]*\)".*/\1/p' "${PLUGIN_MANIFEST}" 2>/dev/null | head -n1)"
+  fi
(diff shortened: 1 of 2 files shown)


kunickiaj:
* **#700** <a href="https://app.graphite.com/github/pr/kunickiaj/codemem/700?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kunickiaj/codemem/700?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#699** <a href="https://app.graphite.com/github/pr/kunickiaj/codemem/699?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

kunickiaj:
### Merge activity

* **Apr 10, 10:54 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/kunickiaj/codemem/700).
* **Apr 10, 10:55 PM UTC**: [Graphite](https://app.graphite.com/github/pr/kunickiaj/codemem/700) rebased this pull request as part of a merge.
* **Apr 10, 10:57 PM UTC**: @kunickiaj merged this pull request with [Graphite](https://app.graphite.com/github/pr/kunickiaj/codemem/700).

PR #923, opened 2026-04-20
fix(ui): honor the Advanced diagnostics Redact toggle

kunickiaj:
## Description

The Redact toggle under Advanced diagnostics was a no-op because every `/api/sync/status` and `/api/sync/pairing` request hardcoded `includeDiagnostics=false` (and `=true` for pairing). Server-side redaction was always applied on status and never on pairing, so flipping the client switch couldn't change what the viewer rendered — there was no unredacted data to reveal on status, and the pairing payload was already raw.

Three coordinated fixes:

1. `api/sync.ts`: `loadPairing()` takes an `includeDiagnostics` flag instead of hardcoding it.
2. `tabs/sync/index.ts`: `loadSyncData` + `loadPairingData` derive the flag from `!isSyncRedactionEnabled()` so both requests match the toggle state. The cache path is untouched (cache is only consulted on the Health tab, which doesn't surface addresses).
3. `diagnostics/lifecycle.ts`: `onCheckedChange` calls the existing `refreshCallback` (wired to `app.ts refresh()`) so flipping the switch immediately re-fetches with the new `includeDiagnostics` value before the rerender runs. Previously the rerender just echoed the prior payload.

## Testing

- pnpm run tsc
- pnpm run lint
- pnpm run test (1417 passed)

## Checklist

- [x] Toggle off → server returns raw addresses + pairing payload + last_error
- [x] Toggle on → server redacts
- [x] Toggle flip triggers refetch before rerender

### DIFF
--- a/packages/ui/src/lib/api/sync.ts
+++ b/packages/ui/src/lib/api/sync.ts
@@ -34,8 +34,9 @@ export async function loadSyncActors(): Promise<unknown> {
 	return fetchJson("/api/sync/actors");
 }
 
-export async function loadPairing(): Promise<unknown> {
-	return fetchJson("/api/sync/pairing?includeDiagnostics=1");
+export async function loadPairing(includeDiagnostics = false): Promise<unknown> {
+	const suffix = includeDiagnostics ? "?includeDiagnostics=1" : "";
+	return fetchJson(`/api/sync/pairing${suffix}`);
 }
 
 export async function updatePeerScope(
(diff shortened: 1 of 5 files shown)


kunickiaj:
### Merge activity

* **Apr 21, 12:05 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/kunickiaj/codemem/923).
* **Apr 21, 12:06 AM UTC**: @kunickiaj merged this pull request with [Graphite](https://app.graphite.com/github/pr/kunickiaj/codemem/923).

PR #993, opened 2026-04-30
fix(core): scan-secrets sweep stack overflow + spurious updates

kunickiaj:
## Description

Two bugs in the retroactive scanner sweep, found running `codemem db scan-secrets --dry-run` against a real 32k-row store:

1. **Stack overflow on large stores.** The sweep accumulated a `ScanDetection[][]` per row and finished with `mergeDetections(...detectionLists)`. Spreading >65k entries into a function call trips V8's argument-stack limit ("Maximum call stack size exceeded"). Reproduces reliably between rows ~10 256 and 10 257 in id order. Fix: replace the array-of-arrays with an inline `Map<kind, count>` accumulator updated as each row is scanned. No spread, constant memory.

2. **Spurious "changed" via JSON canonicalization.** `scanMetadataJson` parsed the stored `metadata_json`, ran `redactValue`, and compared `JSON.stringify(parsed)` against the original string. Any difference in whitespace or key order between the stored JSON and V8's canonical stringify made `nextJson !== json` always true, so every row with metadata reported `changed=true` even when no secret was detected. Confirmed with `--dry-run --json --limit=100`: result was `updated: 100, detections: []`. Fix: gate change-detection on `redactValue`'s actual detections — if none, return the original bytes unchanged.

Also adds **per-row samples** to the result so users can audit hits before running the real sweep. Each affected memory is reported as `{ id, redactedTitle, detections }`. The redacted title is safe to surface (markers, not the original value), and pointing at the row id lets a user `sqlite3` the body to verify whether a hit is a real secret. Surfaced in CLI dry-run output and `--json` mode.

Verified on a real 31 905-row local store: sweep now completes, reports `updated: 9` (real redactions only), and the affected memory IDs let a quick eyeball reveal that several hits are likely rule-tuning false positives (Backstage tokens flagged as AWS keys, Google Doc IDs flagged as 40-char base64 secrets).

## Type of Change

## Testing

`maintenance.test.ts` adds two regressions: one seeds metadata with non-canonical whitespace and asserts a no-secret row is reported `updated=0` / `detections=[]`; the other seeds 12 000 clean rows (above the prior spread cutoff) and asserts the sweep does not throw and reports `updated=0`. Existing sweep tests (legacy redact, idempotency, dry-run, oversize, junction-table refresh, TOCTOU rev-guard) updated to match the new `samples` field on the result.

- [x] Manually verified changes work as expected — full sweep over local 32k-row store completes, 9 of 9 hits surface their memory IDs

## Checklist

- [x] Self-review completed
- [x] Documentation updated (if needed) — JSDoc on the new `ScanSecretsRetroactiveSample` type


kunickiaj:
* **#993** <a href="https://app.graphite.com/github/pr/kunickiaj/codemem/993?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kunickiaj/codemem/993?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

kunickiaj:
### Merge activity

* **Apr 30, 8:40 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/kunickiaj/codemem/993).
* **Apr 30, 8:41 PM UTC**: @kunickiaj merged this pull request with [Graphite](https://app.graphite.com/github/pr/kunickiaj/codemem/993).

PR #1278, opened 2026-06-21
fix(cli): prune-replication-ops prunes all scopes + honest whole-table size report

kunickiaj:
## Description

The offline `codemem db prune-replication-ops` only pruned the **default** replication scope — it called `pruneReplicationOpsUntilCaughtUp` with no `scopeId`. So `replication_ops` accumulated in other scopes (e.g. `legacy-shared-review`, `oss`) was **never reclaimed**: the command was useless for the exact bloated-DB case it exists for, and is a root cause of unbounded `replication_ops` growth. The background retention runner already prunes every scope; the CLI now does too.

- **Prune all scopes:** export `listRetentionScopeIds` (the runner's scope enumerator) from core; the CLI enumerates every scope present in `replication_ops` + the v2 state tables (DEFAULT first) and runs `pruneReplicationOpsUntilCaughtUp` per scope. Each scope's per-scope retained floor still protects peer-needed ops.
- **Honest size report:** the after-size used a per-scope *logical* content estimate while the before-size was the whole-table *physical* dbstat, producing bogus deltas (e.g. "894 MB → 14 MB"). Both now use the same whole-table dbstat (`estimateReplicationOpsBytes`), honestly showing the pre-vacuum size.

## Type of Change

- [ ] 🚀 Feature (new functionality)
- [x] 🐛 Bug fix (fixes an issue)
- [ ] 📚 Documentation (docs-only change)
- [ ] 🔧 Maintenance (refactor, chore, CI, etc.)
- [ ] 🧪 Testing (test-only changes)

## Testing

- [x] Relevant checks pass locally (`pnpm run tsc`, `pnpm run lint`, `pnpm run test`)
- [x] Added/updated tests for changes
- [x] Manually verified changes work as expected

Test: ops seeded in default + `oss` + `legacy-shared-review` (all past the cutoff, no cursors) are all pruned — previously only the default scope cleared. Full `packages/core` + `packages/cli` suites pass (1844).

## Checklist

- [x] Code follows project style (`pnpm run lint` passes for touched files)
- [x] Self-review completed
- [x] Documentation updated (if needed)
- [x] No new warnings introduced

PR #1279, opened 2026-06-21
fix(core): refresh churned replication ops in place to bound op-log growth

kunickiaj:
## Description

`recordReplicationOp` inserted a row into `replication_ops` for every memory save. Anything that re-saved a shared memory on a fixed cadence — bumping `rev`/`updated_at` with no content change — amplified the op log; one memory accumulated 245 byte-identical ops differing only in `rev`/`updated_at`. That bloat is what kept the op log from shrinking after retention pruning.

This adds a content-idempotency guard. When the most recent op for the same `entity_id` + `scope_id` is the same `op_type` and carries identical memory content (volatile `rev`/`updated_at`/`clock_*` fields ignored), the existing op is **refreshed in place** instead of inserting a duplicate row.

Key invariants preserved:

- **Cursor-stable.** The refresh reuses the existing `op_id` and `created_at`, so the op's position in the outbound feed (`loadReplicationOpsSince`, cursored on `created_at|op_id`) is unchanged — already-synced peers don't re-pull.
- **`clock_rev == rev` holds.** The refresh still advances `clock_rev` to the memory's current `rev`. `hasUnsyncedSharedMemoryChanges` and `backfillReplicationOps` both treat a shared memory as needing an op whenever no op has `clock_rev == rev`; a frozen `clock_rev` would make a content-stable memory read dirty forever and abort sync.
- **Resurrection still emits.** The lookup is not filtered by `op_type`, so an `upsert` after a `delete` still emits even when its content matches a pre-delete `upsert`.
- **No content-eater.** Volatile fields are matched against an explicit allow-set at every nesting level — no `clock_*` prefix wildcard that could silently drop a future content field.

Out of scope: whatever upstream save path bumps `rev` on a no-op write (the churn trigger itself) — this bounds the op-log symptom regardless.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Testing

- `pnpm run tsc` — clean
- `pnpm run lint` — clean
- `pnpm exec vitest run packages/core` — 1577 passing (incl. new churn-dedup, dirty-check, resurrection, cross-scope-delete, and `replicationOpContentEqual` cases)
- `pnpm exec vitest run packages/cli` — 278 passing

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code in hard-to-understand areas
- [x] I have added tests that prove my fix is effective
- [x] New and existing unit tests pass locally with my changes

### DIFF
--- a/packages/core/src/sync-replication.test.ts
+++ b/packages/core/src/sync-replication.test.ts
@@ -29,6 +29,7 @@ import {
 	reconcileStalePeerReceivedRows,
 	recordAccessCleanupOp,
 	recordReplicationOp,
+	replicationOpContentEqual,
 	SCOPED_NULL_BASELINE_BOOTSTRAP_CURSOR_MARKER,
 	setReplicationCursor,
 	setSyncResetState,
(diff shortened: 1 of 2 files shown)


kunickiaj:
* **#1279** <a href="https://app.graphite.com/github/pr/kunickiaj/codemem/1279?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kunickiaj/codemem/1279?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

kunickiaj:
Closing: this in-place op-refresh can't be made correct at the op layer.

Advancing clock_rev on a content-identical refresh is required so the shared-memory dirty-check (clock_rev == rev) doesn't read dirty forever and block stale-cursor auto-bootstrap — but it also hides the rev advance from peers that already passed the op's `created_at|op_id` cursor. A peer that then makes a real lower-rev edit gets LWW-rejected against the locally-churned higher rev → non-convergence. The dirty-check and LWW impose contradictory constraints on clock_rev whenever rev churns, so the real fix has to prevent the no-op rev bump at the source, not dedup after the fact.

Investigation found no active rev-churn trigger in the current code: every memory_items.rev write is a deliberate mutation or an inbound apply, and the op-amplification incident this targeted was historical (peer device, ~120s cadence, stopped late May). The bloat it left is reclaimed by #1278 (prune all scopes + honest size). Worth revisiting only with fresh evidence of a live churn trigger, at which point the correct fix guards that specific write path.

PR #1319, opened 2026-07-09
fix(plugin): suppress execSync stderr noise on OpenCode startup

### LINKED ISSUE #1282: Harmless stderr noise on startup: 'codemem: not found' and 'fatal: not a git repository' leak from execSync
## Problem

When OpenCode starts with the codemem plugin installed via `npx -y codemem setup --opencode-only` (no global install), two error messages leak to the terminal:

```
/bin/sh: 1: codemem: not found
fatal: not a git repository (or any of the parent directories): .git
```

Both are **harmless** — the plugin catches both via try/catch and continues normally. However, the stderr output from `execSync` is not suppressed, so it leaks to the user's terminal and looks like something is broken.

### 1. `codemem: not found` — `detectRunner()` at line 625

```js
execSync("codemem --version", { encoding: "utf-8", timeout: 3000 }).trim()
```

The plugin checks if `codemem` is globally installed. When it's not on PATH (common with `npx`-only setups), the shell prints `codemem: not found` to stderr. The failure is caught at line 644 and falls back to `npx`, but the stderr has already leaked.

### 2. `fatal: not a git repository` — plugin init at line 1651

```js
execSync("git rev-parse --short HEAD", { cwd: runnerFrom, timeout: 500, encoding: "utf-8" }).trim()
```

The plugin tries to get a git commit hash for debug logging. When the working directory is not a git repo, git prints the fatal error to stderr. The exception is caught at line 1656 and `version` is set to `"unknown"`, but again the stderr has already leaked.

## Suggested Fix

Both `execSync` calls should suppress stderr by adding `stdio: pipe` (or switching to `execFileSync` which pipes stderr by default):

```js
// Line 625
execSync("codemem --version", { encoding: "utf-8", timeout: 3000, stdio: pipe }).trim()

// Line 1651
execSync("git rev-parse --short HEAD", { cwd: runnerFrom, timeout: 500, encoding: "utf-8", stdio: pipe }).trim()
```

This keeps the existing try/catch behavior but prevents the noise from reaching the terminal.

## Environment

- **Plugin**: `@codemem/opencode-plugin@0.37.0`
- **Install method**: `npx -y codemem setup --opencode-only` (no global install)
- **OpenCode**: running in a non-git working directory
- **OS**: Linux

## Workaround

Users can silence these by:
- `npm install -g codemem` (puts codemem on PATH) + `git init` in the working directory
- Or set `CODEMEM_RUNNER=npx` env var to skip global detection, and ignore the git error

ZeR020:
## Summary

- Pipe `stdio` for two `execSync` probes in the OpenCode plugin so expected failures no longer leak shell/git stderr to the terminal on every startup.
- `detectRunner()`: `codemem --version` when the CLI is not on PATH (npx-only installs) still falls back to `npx`.
- Plugin init: `git rev-parse --short HEAD` in non-git working directories still leaves `version` as `"unknown"`.

Fixes #1282

## Problem

On OpenCode startup with `@codemem/opencode-plugin` (npx-only install, non-git cwd), users see:

```
/bin/sh: 1: codemem: not found
fatal: not a git repository (or any of the parent directories): .git
```

Both failures are already caught; only stderr was leaking.

## Change

```js
// detectRunner()
execSync("codemem --version", {
  encoding: "utf-8",
  timeout: 3000,
  stdio: ["pipe", "pipe", "pipe"],
})

// plugin init
execSync("git rev-parse --short HEAD", {
  cwd: runnerFrom,
  timeout: 500,
  encoding: "utf-8",
  stdio: ["pipe", "pipe", "pipe"],
})
```

No behavior change — only suppresses cosmetic noise.

## Test plan

- [x] Confirmed both call sites still present and unpiped on `main` / 0.39.0 before the change
- [x] Local Node repro: `execSync` with `stdio: ["pipe","pipe","pipe"]` for a missing command and `git rev-parse` in `/tmp` catches without printing stderr
- [ ] Start OpenCode with npx-only codemem install in a non-git directory — no `codemem: not found` / `fatal: not a git repository` noise
- [ ] With global `codemem` on PATH, `detectRunner()` still prefers the global binary when version matches pin
- [ ] In a git repo, plugin init still logs a short commit hash as `version`

### DIFF
--- a/packages/opencode-plugin/.opencode/plugins/codemem.js
+++ b/packages/opencode-plugin/.opencode/plugins/codemem.js
@@ -892,7 +892,7 @@ const detectRunner = ({ cwd, envRunner }) => {
       // Suppress shell "not found" noise when codemem is not on PATH; the
       // catch below falls back to npx. Without this, stderr leaks to the
       // terminal on every OpenCode startup for npx-only installs.
-      stdio: ["pipe", "pipe", "pipe"],
+      stdio: ["ignore", "pipe", "ignore"],
     }).trim();
     // Accept any global codemem at or above the pinned backend version, so a
     // newer global install is used instead of silently downgrading to
@@ -1955,7 +1955,7 @@ export const OpencodeMemPlugin = async ({
       encoding: "utf-8",
       // Suppress "fatal: not a git repository" when the working directory is
       // not a git repo. The catch below leaves version as "unknown".
-      stdio: ["pipe", "pipe", "pipe"],
+      stdio: ["ignore", "pipe", "ignore"],
     }).trim();
   } catch (err) {
     // Ignore - version will remain 'unknown'
--- a/packages/opencode-plugin/.opencode/plugins/codemem.js
+++ b/packages/opencode-plugin/.opencode/plugins/codemem.js
@@ -886,7 +886,14 @@ const detectRunner = ({ cwd, envRunner }) => {
   }
   // Prefer the TS codemem if installed globally, fall back to npx
   try {
-    const versionOutput = execSync("codemem --version", { encoding: "utf-8", timeout: 3000 }).trim();
+    const versionOutput = execSync("codemem --version", {
+      encoding: "utf-8",
+      timeout: 3000,
+      // Suppress shell "not found" noise when codemem is not on PATH; the
+      // catch below falls back to npx. Without this, stderr leaks to the
+      // terminal on every OpenCode startup for npx-only installs.
+      stdio: ["pipe", "pipe", "pipe"],
+    }).trim();
     // Accept any global codemem at or above the pinned backend version, so a
     // newer global install is used instead of silently downgrading to
     // `npx codemem@<pin>` (the old check only matched the exact pin or the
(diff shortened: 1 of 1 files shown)


kunickiaj (dismissed):
One small thing and I think this is good to merge:

1. stdio intent — ["pipe", "pipe", "pipe"] works, but since we never write to the child's stdin and we discard stderr in the catch, ["ignore", "pipe", "ignore"] says what we actually mean and avoids allocating pipes we throw away. Totally optional, but it reads more clearly as "suppress this."

Nothing else from me. Nice contribution. 🙏

kunickiaj (approved):

kunickiaj:
### Merge activity

* **Jul 13, 6:45 AM UTC**: [Graphite](https://app.graphite.com/github/pr/kunickiaj/codemem/1319) couldn't merge this PR because it failed for an unknown reason (Fast-forward merges are not supported for forked repositories. Please create a branch in the target repository in order to merge).

ZeR020:
Hey @kunickiaj 👋 — the Graphite fast-forward merge fails on forked repos (GitHub limitation). The regular GitHub "Merge pull request" button works fine for fork PRs though. No code changes needed — just the merge method. Thanks!

PR #1430, opened 2026-08-10
feat: add pi agent support (@codemem/pi-extension)

### LINKED ISSUE #1429: feat: add pi agent support
## Summary

Add first-class support for the [pi](https://github.com/earendil-works/pi-coding-agent) coding agent as codemem's fourth client integration, alongside opencode, Claude Code, and Codex.

## Motivation

Pi is a deliberately minimal, TypeScript-native coding agent whose only plugin surface is its extension API (`pi.on()` events + `pi.registerTool()`); it has **no native MCP support and no hooks mechanism** — extensions loaded from npm/git/local "pi-packages" are the entire integration surface. Its extension API is arguably the richest of the four clients: per-turn `before_agent_start` handlers can append to the system prompt, `tool_result` middleware can patch results, and `session_before_compact` exposes compaction boundaries.

A pi integration built on codemem's existing thin-client pattern (HTTP to the viewer server + CLI fallback) lands at opencode-level first-classness, plus two pi-only capabilities:

- **compaction-boundary memory extraction** — flush extraction before pi discards context (observe-only; never replaces pi's own summarizer)
- **fork/resume-aware raw-event streams** — pi session ids mint per fork

## Proposed scope

- **`@codemem/pi-extension`** (new npm pi-package, lockstep-versioned with `@codemem/opencode-plugin`): event ingest via `POST /api/pi-hooks`, per-turn memory injection via system-prompt append (never the persistent message channel), and codemem's 14 `memory_*` tools registered **natively** (no pi-mcp-adapter required; optional `--pi-mcp` opt-in for adapter users).
- **core**: pi event adapter (AdapterEvent v1 with `source: "pi"`, deterministic event ids, flush-only compaction signals) + observer extraction configuration derived from pi's own `settings.json`/`models.json`/`auth.json` (API-key providers only in v1; OAuth reported as explicitly unsupported rather than silently 401).
- **viewer-server**: dedicated `POST /api/pi-hooks` (server-side source assertion) + closes the HTTP route gap so all 14 tools work over HTTP.
- **cli**: `pi-hook-ingest` (HTTP-first, spool + boundary flush) / `pi-hook-inject` commands, and `codemem setup` pi target — one-command install (`--pi-only`, `--pi-mcp`, `--pi-extension-path`).
- **ui/docs**: pi in settings/health client enumerations; README + docs updated in the same PR.

Because the store is shared, pi users who also run opencode/claude/codex get cross-agent memory automatically (pack filtering is project-scoped, never agent-scoped).

## Non-goals (this iteration)

- OAuth-based extraction via pi providers (API-key providers only in v1).
- Google / non-OpenAI-or-Anthropic observer wire APIs.
- Replacing or customizing pi's compaction summarizer (observe-only).
- Generalizing `/api/raw-events` to arbitrary sources (dedicated route instead, mirroring `/api/claude-hooks`).

I have this implemented and reviewed (multi-pass adversarial review + full quality gates green) and will open the linked PR shortly.

ZeR020:
Closes #1429

## Summary

First-class support for the [pi](https://github.com/earendil-works/pi-coding-agent) coding agent, at the same depth as the opencode integration — plus two pi-only capabilities (compaction-boundary extraction, fork-aware streams).

pi has no native MCP or hooks mechanism; its extension API *is* the hook surface. This PR adds the fourth client integration following the existing AdapterEvent → raw-events → sweeper → ingest pipeline:

- **`@codemem/pi-extension`** (new npm pi-package): thin client — event ingest, per-turn system-prompt memory injection, **all 14 `memory_*` tools registered natively** (no third-party MCP adapter required). Zero native deps inside pi's process (dist imports node builtins only).
- **core**: pi event adapter mirroring `claude-hooks` (`source: "pi"`, deterministic `pi:<session>:<entry>` ids, flush-only compaction signals) + pi-derived observer config (API-key providers only in v1; OAuth explicitly unsupported-with-status; cheap-model-first; credentials resolved in-memory at point of use, never persisted/logged).
- **viewer-server**: dedicated `POST /api/pi-hooks` (server-side source assertion) + closes the HTTP route gap so all 14 tools work over HTTP (`remember`, `timeline`, `expand`, `schema`, `search_index`, `explain`, `distill_candidates`).
- **cli**: `pi-hook-ingest` (+spool, boundary flush) / `pi-hook-inject`, and `codemem setup` pi target (`--pi-only`, `--pi-mcp` adapter opt-in, `--pi-extension-path` dev mode; idempotent; upgrades stale npm pins without `--force`).
- **ui/docs**: pi in settings/health client enumerations; README + docs updated in the same PR.

Shared store ⟹ multi-agent users get cross-agent memory (pack filters are project-scoped, never agent-scoped).

## Attribution hardening

~20 sites silently default an event's source to `"opencode"`; every pi path passes `source: "pi"` explicitly and the ingestion route asserts it server-side. Tests assert pi-ingested rows attribute to pi and create **zero** opencode rows (store + route + CLI levels).

## Test plan

- New packages covered: 38 (adapter) + 21 (observer config) + 15 (routes) + 33 (extension) + 62 (setup/cli) tests — every regression from review has a test that fails against the unfixed code
- `pnpm run lint && pnpm run tsc && pnpm run test && pnpm build` green
- CONTRIBUTING injection suites: pack/pack.eval/memory-inject/claude-hooks **168/168**; opencode plugin `test:plugin` **122/122** (no regressions for existing clients)
- Live dogfood: pi events ingested with `source=pi` (0 opencode rows), retry dedupe, remember→inject round-trip, `## codemem memories` block
- Known pre-existing flakes: `raw-event-sweeper`/`raw-event-flush`/`ingest-pipeline` timeout-sensitive tests can fail under heavy load (they pass 179/179 in isolation) — same class exists on `main`, untouched by this PR

## Known limitations (documented)

- Observer extraction from pi config supports **API-key providers** only in v1; OAuth-only installs report `unconfigured (oauth-only)` with guidance instead of silent 401s
- `pi-hook-ingest` lock-busy boundary-flush gap in the **claude** twin is pre-existing upstream behavior; pi's twin handles all paths — the claude fix is left for a follow-up PR
- `codemem setup`'s settings.json write is JSON (JSONC comments in *pi's* file are not preserved; `.bak` backup is written first — same behavior as the existing opencode.jsonc handling)


packages/viewer-server/src/routes/raw-events.ts
```diff
@@ -483,5 +484,53 @@ export function rawEventsRoutes(getStore: StoreFactory, sweeper?: RawEventSweepe
...
+			store.updateRawEventSessionMeta({
+				opencodeSessionId,
+				source,
+				cwd: envelope.cwd,
+				project: envelope.project,
+				startedAt: envelope.started_at,
+				lastSeenTsWallMs: envelope.ts_wall_ms,
+			});
+
+			nudgeSweeper(sweeper, [opencodeSessionId], source);
```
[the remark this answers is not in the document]
ZeR020: Valid observation, applied: the local binding is now named sessionStreamId and the store API field keeps its opencodeSessionId name explicitly, so the pi route no longer implies an opencode origin. Commit aee5090a.

packages/pi-extension/src/tools.ts
```diff
@@ -0,0 +1,889 @@
...
+
+function parseCliJson(stdout: string): unknown {
+	const trimmed = stdout.trim();
+	if (!trimmed) return null;
+	// CLI may emit logs before JSON — take the last {...} or [...] block.
+	const objStart = trimmed.lastIndexOf("{");
+	const arrStart = trimmed.lastIndexOf("[");
+	const start = Math.max(objStart, arrStart);
+	if (start < 0) return trimmed;
+	return JSON.parse(trimmed.slice(start));
```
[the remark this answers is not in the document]
ZeR020: Confirmed and fixed: parseCliJson now scans candidate JSON value starts in document order and returns the first suffix that parses, so nested shapes like the items-with-inner-object example work with viewer down. Added regression tests for the exact nested-object case plus logs-before-JSON. Commit aee5090a.

packages/pi-extension/src/tools.ts
```diff
@@ -0,0 +1,889 @@
...
+			}),
+			async execute(_id, rawParams, signal) {
+				const params = paramsOf(rawParams);
+				return withToolError("memory_distill_candidates", async () => {
+					const body = {
+						...(params as Record<string, unknown>),
+						project:
+							projectOrClient(params as Record<string, unknown>, client) ??
+							(params as { project?: string }).project,
+					};
```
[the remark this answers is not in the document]
ZeR020: Verified against the copy in memory-tools.ts:332, which rejects project combined with all_projects, so this was a real failure mode when client.project was set. buildDistillBody now drops project whenever all_projects is requested, on both the HTTP body and the CLI fallback args. Regression tests cover client.project being set. Commit aee5090a.

ZeR020:
@kunickiaj Could you please take a look at #1430 when you have a moment? It has been open since August 10; latest `main` is merged and CI is green. Happy to address any feedback. Thank you.

packages/core/src/observer-client.ts
```diff
@@ -1493,6 +1555,31 @@ export class ObserverClient {
 		});
 		this.auth = { token: null, authType: "none", source: "none" };
 
+		// D8: resolve pi credential in memory at point of use. Never assign onto
+		// cfg.observerApiKey (that would look "explicit" and could be persisted
+		// by callers of toConfig()). Only used as a lower-priority cascade source.
+		if (!this._apiKey) {
+			this._piApiKey = resolvePiApiKeyForObserver();
```
kunickiaj, line 1562: **Blocking:** `resolvePiApiKeyForObserver()` is not provider-scoped. `_piApiKey` is subsequently offered to the custom, Anthropic, and OpenAI auth branches. If pi selects an API key for provider A while codemem explicitly configures provider B, provider A’s key can be transmitted to provider B’s endpoint.

Please resolve the credential for the effective observer provider and return it only when the pi provider matches. Add a regression test with an explicit Anthropic/OpenAI observer and an unrelated pi provider key.

packages/pi-extension/src/index.ts
```diff
@@ -0,0 +1,526 @@
...
+		await safeIngest(client, payload, state, pi, ctx.signal);
+
+		// File-context: attach relevant memories when read completes (fail-open).
+		if (config.fileContext && toolName === "read" && !event.isError) {
+			const path = readPathFromToolInput(toolInput);
+			if (path) {
+				try {
+					const block = await fetchFileContextBlock(client, path, ctx.signal);
+					if (block) {
+						const existing = Array.isArray(event.content) ? [...event.content] : [];
```
kunickiaj, line 406: **Blocking:** when `event.content` is a string, `existing` becomes `[]`, so the read result is replaced by the memory block. This package’s serializer already handles string content, and pi’s tool result contract permits text content.

Normalize string content into a text item before appending, and add a regression test asserting that the original file contents remain present.

packages/pi-extension/src/client.ts
```diff
@@ -0,0 +1,386 @@
...
+
+	private async tryCliIngest(
+		payload: Record<string, unknown>,
+		signal?: AbortSignal,
+	): Promise<IngestOutcome> {
+		try {
+			const { stdout } = await this.execCodemem(["pi-hook-ingest"], {
+				stdin: JSON.stringify(payload),
+				signal,
+				timeoutMs: this.config.httpTimeoutMs + 2000,
```
kunickiaj, line 190: **Blocking:** boundary events always use this CLI path, but the default timeout is only seven seconds. The command may perform HTTP retries, lock acquisition, spool draining, and `flushRawEvents`, including an observer LLM request. It can therefore receive `SIGTERM` during the extraction that shutdown/compaction is specifically intended to preserve.

Boundary events need a substantially larger or independently configured timeout, plus a test demonstrating that a long-running flush is not killed.

packages/pi-extension/src/index.ts
```diff
@@ -0,0 +1,526 @@
...
+		if (!config.injectPrompts) return;
+		try {
+			const prompt = typeof event.prompt === "string" ? event.prompt : "";
+			if (!prompt.trim()) return;
+
+			const packOrBlock = await client.fetchPackText(prompt, ctx.signal);
+			if (!packOrBlock.trim()) return;
+
+			// pi-hook-inject already returns the full ## block; HTTP pack is bare pack_text.
+			const block = packOrBlock.includes("## codemem memories")
```
kunickiaj, line 449: **Blocking:** this treats a substring inside recalled memory content as proof that the transport returned a preformatted block. A memory that happens to contain `## codemem memories` bypasses `formatPiInjectionBlock()` and therefore the reference-data safety framing.

Please have `fetchPackText()` return a discriminated result such as `{ text, preformatted }` instead of inferring transport state from content.

packages/pi-extension/src/client.ts
```diff
@@ -0,0 +1,386 @@
...
+				child.stdin.write(opts.stdin);
+				child.stdin.end();
+			} else if (child.stdin) {
+				child.stdin.end();
+			}
+		});
+	}
+
+	/** Best-effort project label from cwd basename (viewer accepts explicit project). */
+	static projectFromCwd(cwd: string): string | null {
```
kunickiaj, line 379: **Blocking:** ingestion resolves the project from the Git root, but this derives retrieval scope from the cwd basename. Running pi from `repo/packages/api` can write memories under `repo` and query them under `api`, producing silently empty injection and tool results.

Please match core’s project resolution semantics, including walking to the repository root, and add a nested-cwd regression test.

packages/cli/src/commands/pi-hook-ingest.ts
```diff
@@ -0,0 +1,472 @@
...
+			logHookEvent(
+				`codemem pi-hook-ingest backlog drain failed: ${err instanceof Error ? err.message : String(err)}`,
+			);
+		}
+	};
+
+	// 1. Unlocked HTTP attempt — fast path when the viewer is up.
+	const httpResult = await httpIngest(payload, opts.host, port);
+	if (httpResult.ok) {
+		await flushOnBoundaryIfRequested();
```
kunickiaj, line 358: **Blocking:** the successful HTTP path flushes before draining the backlog. For a shutdown/compaction event, spooled earlier events are inserted only after the final extraction, leaving the session tail unprocessed.

Drain the backlog before `flushOnBoundaryIfRequested()`. The locked retry path already uses the correct ordering.

packages/pi-extension/package.json
```diff
@@ -0,0 +1,69 @@
+{
+	"name": "@codemem/pi-extension",
+	"version": "0.40.1",
```
kunickiaj, line 3: `pnpm run release:version -- check` currently fails because this package is `0.40.1` while the managed release set is `0.40.2`. Please align it before the release-wiring layer lands.

packages/viewer-server/src/routes/memory-tools.ts
```diff
@@ -0,0 +1,770 @@
...
+
+export function memoryToolRoutes(getStore: StoreFactory) {
+	const app = new Hono();
+
+	// POST /api/memories/remember — twin of memory_remember
+	app.post("/api/memories/remember", async (c) => {
+		const store = getStore();
+		let body: unknown;
+		try {
+			body = await c.req.json();
```
kunickiaj, line 494: These new POST routes use unbounded `c.req.json()` even though `raw-events.ts` already has a bounded JSON parser. Please reuse the bounded parser for `remember`, `expand`, `explain`, and `distill_candidates`, and validate filter value types before passing them to SQLite.

kunickiaj (changes requested):
Thanks for putting substantial work into this. I’m open to supporting pi, especially if you’re willing to remain the primary maintainer for the integration.

I can’t merge this in its current form, though. This is not only about the 12.6k-line size: review found several correctness and credential-handling blockers, detailed inline.

The main blockers are:

- A pi API key can be offered to a different observer provider, potentially sending one provider’s credential to another provider’s endpoint.
- File-context enrichment can replace string tool output instead of appending to it.
- Compaction/shutdown extraction receives a seven-second timeout despite performing retries, spool draining, and an LLM call.
- Injection framing is detected using a substring of untrusted memory content.
- Project resolution uses the cwd basename while ingestion uses the Git root, silently breaking retrieval in monorepos and subdirectories.
- The successful boundary path flushes before draining the spooled backlog.
- The new package version is already out of alignment with the release set.

There are also some important follow-ups:

- The new memory-tool POST routes should reuse the bounded JSON parser and validate filter types.
- The spool/locking implementation should be shared rather than adding a third copy of concurrency-sensitive code.
- MCP/viewer memory access and schema behavior should share core primitives to avoid authorization and contract drift.
- The retrieval-ledger behavior for pi’s preferred HTTP path needs an explicit decision.

I’m also asking that this be converted into a stack rather than continuing as one PR. The existing commits already roughly follow suitable boundaries, so this should preserve the work rather than require starting over. A reasonable stack would be:

1. Shared spool/memory-access refactors
2. Core pi event adapter
3. Viewer routes
4. Pi observer configuration/auth
5. CLI ingest/inject
6. Pi extension lifecycle/client
7. Native tools
8. Setup, UI, docs, and release wiring

Graphite or GitHub’s native stacked-PR workflow is fine. Each layer should remain independently reviewable and testable.

Local validation at `898ccbdd`:

- Build passed
- TypeScript passed
- Biome passed
- Targeted tests passed: 292/292 after generating viewer assets
- `pnpm run release:version -- check` failed because `@codemem/pi-extension` is `0.40.1` while the release set is `0.40.2`

Once the stack is created and the blocking findings are addressed, I’m happy to review it layer by layer.

ZeR020:
@kunickiaj Thank you for the review.

I split the work into the eight stacked PRs you asked for, rebased onto current main (0.41). This PR stays open; it is not the merge path.

Please start at #1473 (layer 1; mergeable). Layers 2–8 are on the fork for now, because GitHub cannot base a PR on `kunickiaj:stack/…`:

- layer 2: https://github.com/ZeR020/codemem/pull/1
- layer 3: https://github.com/ZeR020/codemem/pull/2
- layer 4: https://github.com/ZeR020/codemem/pull/3
- layer 5: https://github.com/ZeR020/codemem/pull/4
- layer 6: https://github.com/ZeR020/codemem/pull/5
- layer 7: https://github.com/ZeR020/codemem/pull/6
- layer 8: https://github.com/ZeR020/codemem/pull/7 (`Closes #1429`)

After #1473 merges I will retarget layer 2 at this repo’s `main`, so each later layer lands here one at a time.

The seven blockers are fixed in the layer that owns them, with tests. I will do what I can on the pi side and fix issues or bugs in this integration. I cannot guarantee I will maintain it forever.

Happy to walk through any layer.

kunickiaj (approved):

kunickiaj:
Closing as superseded by the eight reviewed and merged Pi slices: #1473, #1546, #1678, #1693, #1744, #1767, #1778, and #1788. Pi support is now on `main`; no code from this umbrella PR remains to merge.

Release follow-up: bootstrap `@codemem/pi-extension` on npm, then configure its trusted publisher before the first release tag that includes it.

PR #1473, opened 2026-08-17
refactor: share hook spool, bounded JSON parser, and memory kinds

ZeR020:
First stacked PR for the pi integration review on #1430. No pi client code in this layer.

## Why

#1430 asked that spool/locking be shared instead of adding a third copy, that new memory-tool POSTs reuse the bounded JSON parser, and that MCP/viewer memory kinds share a core catalog. This PR extracts those primitives on current `main` so later pi layers can sit on them.

## What

- Parameterized `createHookIngestSpool` used by claude and codex wrappers (same export names, same TTL/attempt numbers)
- Claude HTTP-success path drains the backlog **before** the boundary flush
- `parseJsonObjectBody` exported from `packages/viewer-server/src/helpers.ts`
- Core `memory-kinds.ts`: seven remember kinds + `session_summary`; MCP re-exports the seven

## Test plan

- `vitest run` on claude-hook-ingest-spool, claude-hook-ingest, codex-hook-ingest, helpers, memory-kinds — 63/63 passed
- `tsc -b packages/cli --force` passed

Stacked follow-ups (not in this PR): core pi adapter → `/api/pi-hooks` alias → observer auth → CLI → extension → tools → setup/docs.

Related: #1430, #1429


packages/core/src/memory-kinds.ts
```diff
@@ -0,0 +1,34 @@
+/**
+ * Canonical memory-kind catalog. The seven kinds below are the ones the
+ * MCP remember tools accept; the store additionally allows
+ * `session_summary` (written by the observer pipeline). Every surface
+ * (MCP schema, store validation, viewer routes) imports this catalog so
+ * the kinds never drift.
+ */
+
+export const MEMORY_KIND_DESCRIPTIONS: Record<string, string> = {
```
kunickiaj: Nice direction—having the descriptions and store validation share one catalog removes an existing source of drift.

Could we also derive the actual MCP validator from this catalog? `packages/mcp-server/src/schemas.ts` still has a separate hardcoded `z.enum(...)`, while `ingest-pipeline.ts` and `ingest-xml-parser.ts` retain similar lists.

Preserving literal keys here with something like `as const satisfies Record<string, string>` would let `REMEMBER_MEMORY_KINDS` drive the MCP schema and make the “canonical catalog” invariant enforceable rather than documentary.

packages/cli/src/commands/codex-hook-ingest-spool.ts
```diff
@@ -1,342 +1,34 @@
...
+ * quarantine machinery is shared in `hook-ingest-spool.ts`; this file
+ * wires the codex-specific config (dirs, TTL 120s, 20 acquire attempts,
+ * error name/message).
+ */
+
+import { createHookIngestSpool } from "./hook-ingest-spool.js";
+
+export type { SpoolDrainResult } from "./hook-ingest-spool.js";
+
+const spool = createHookIngestSpool({
```
kunickiaj, line 12: Could we add a small focused test for the Codex factory configuration? The extraction moves several important behaviors into this config object—environment variable names, TTL, spool directory, retry count, and error identity—but the detailed spool tests currently exercise only the Claude instance.

At minimum, it would be useful to assert the Codex TTL/env overrides and that Claude and Codex resolve to separate spool directories. That would protect against a future configuration typo silently mixing the queues.

packages/cli/src/commands/claude-hook-ingest.ts
```diff
@@ -308,8 +308,11 @@ export async function ingestClaudeHookPayload(
 	// 1. Unlocked HTTP attempt — fast path when the viewer is up.
 	const httpResult = await httpIngest(payload, opts.host, port);
 	if (httpResult.ok) {
-		await flushOnBoundaryIfRequested();
+		// Drain any spooled backlog before the boundary flush so the
+		// flush pass sees every queued payload of the session, not just
+		// this event.
 		await drainBacklogIfPresent();
```
kunickiaj, line 334: Draining the backlog before the boundary flush makes sense and fixes the ordering issue described in the PR.

Could the regression test record both the queued HTTP handling and `boundaryFlush` in one sequence, then assert that the drain completes first? The current test confirms that the backlog drains, but its payload is a `Stop` event with flushing disabled by default, so it does not exercise the ordering this change is intended to guarantee.

packages/cli/src/commands/hook-ingest-spool.ts
```diff
@@ -0,0 +1,507 @@
...
+				// Permanently corrupt content — keeping the file around would
+				// loop forever every drain. Quarantine it under a `.bad-` prefix
+				// so an operator can inspect it without it being picked up again.
+				quarantineSpoolEntry(dir, name, "parse-error");
+				result.failed++;
+				continue;
+			}
+			if (parsed == null || typeof parsed !== "object" || Array.isArray(parsed)) {
+				// Parseable but wrong shape — same problem, same fix.
+				quarantineSpoolEntry(dir, name, "wrong-shape");
```
kunickiaj, line 470: This shared implementation now increments `failed` for wrong-shape Claude spool entries, whereas the previous Claude implementation quarantined them without incrementing the count. The new behavior looks more consistent, but could we assert the returned `{ processed, failed }` result in the existing wrong-shape test so the intentional change is captured?

kunickiaj (changes requested):
Thanks for splitting this foundation out of #1430—the shared spool factory and bounded JSON helper are much easier to review in this form, and the targeted tests pass locally (187 tests plus the CLI TypeScript build).

I found two items I think we should address before merging:

1. This branch currently conflicts with `main`. Recent changes added `rawEventTarget`/`targetMismatch` handling in the same Claude and Codex ingest paths touched here. Could you restack and preserve those safeguards during conflict resolution?
2. The new memory-kind catalog does not yet drive the MCP validation schema, so kind enforcement can still drift.

I also left a couple of focused test suggestions for the new abstraction. Once the branch is restacked and those catalog changes are addressed, I would be happy to take another pass.

ZeR020:
Restacked onto current `main` and kept the `rawEventTarget` / `targetMismatch` path. `REMEMBER_MEMORY_KINDS` now drives the MCP `z.enum` plus the ingest-pipeline and observer XML kind sets.

packages/core/src/memory-kinds.ts
ZeR020: `memoryKindSchema` is now `z.enum(REMEMBER_MEMORY_KINDS)`. Ingest-pipeline and `SUPPORTED_OBSERVATION_KINDS` use the same tuple.

packages/cli/src/commands/codex-hook-ingest-spool.ts
ZeR020: Added `codex-hook-ingest-spool.test.ts` for the 120s TTL, env override, separate spool dirs, and `CodexHookLockBusyError`.

packages/cli/src/commands/claude-hook-ingest.ts
ZeR020: The SessionEnd case already records `http:queued` before `flush:fresh`. Left that test in place.

packages/cli/src/commands/hook-ingest-spool.ts
ZeR020: The existing non-object test now asserts `{ processed: 0, failed: 2 }`.

kunickiaj (approved):
Re-reviewed at f8c95d1. All prior requested changes are addressed: the branch is cleanly restacked with viewer-target safeguards preserved, MCP and observer enforcement now derive from the core memory-kind catalog, Codex spool configuration/isolation has focused coverage, the boundary-flush ordering is asserted, and wrong-shape drain accounting is covered.

Local validation passed: 202 targeted tests, workspace TypeScript build, and Biome lint. CI is also green. No merge blockers remain.

Approving this revision. Please hold the merge until after the 0.43 release so it lands in the intended release sequence.

ZeR020:
@kunickiaj 0.43 is out. This is still approved from Aug 25. Restacked onto current `main`.

kunickiaj:
### Merge activity

* **Aug 28, 7:08 PM UTC**: [Graphite](https://app.graphite.com/github/pr/kunickiaj/codemem/1473) couldn't merge this PR because it failed for an unknown reason (Fast-forward merges are not supported for forked repositories. Please create a branch in the target repository in order to merge).

ZeR020:
@kunickiaj Graphite couldn’t merge this: fast-forward merge isn’t supported for fork PRs. GitHub shows it mergeable with green CI. Could you merge it from the GitHub UI (Merge or Squash) instead of Graphite?

PR #1546, opened 2026-08-29
feat(core): add pi event adapter for 0.41 ingest

ZeR020:
Second PR in the pi integration stack — context and overall design in #1430. Builds on #1473 (merged); the diff here is the pi adapter only.

## Why

Pi events need a core adapter that produces the same envelope shape `ingestRawEvents` already accepts, with `source: "pi"` explicit and deterministic ids so HTTP/CLI retries do not duplicate rows.

## What

- `packages/core/src/pi-hooks.ts`: map session/prompt/tool events; compaction is flush-only
- Envelope fields include `session_stream_id` / `session_id` aliases required by 0.41 ingest
- Test: `ingestRawEvents` persists source `pi` and creates zero opencode rows

## Test plan

- `vitest run packages/core/src/pi-hooks.test.ts` — 40/40 passed
- Gates `lint`, `tsc`, `build` green locally; `test` passes except local-sandbox env failures unrelated to this change (sqlite-vec native ext unavailable in this dev sandbox)

@kunickiaj layer 2 of the stack is ready for review whenever you have a moment. Layers 3–8 will follow one at a time as each predecessor merges, same as this one.

Related: #1430, #1429, #1473

### DIFF
--- a/packages/core/src/pi-hooks.test.ts
+++ b/packages/core/src/pi-hooks.test.ts
@@ -1,0 +1,776 @@
+/**
+ * Tests for pi-hooks.ts — AdapterEvent v1 mapping for pi extension events.
+ *
+ * Covers:
+ *  - mapPiEventPayload: all mappable event types, skip cases, deterministic ids,
+ *    tool_result isError, fork → new stream identity
+ *  - buildPiFlushSignalFromEvent: session_before_compact flush-only contract
+ *  - buildRawEventEnvelopeFromPiEvent: envelope shape + source "pi"
+ *  - buildIngestPayloadFromPiEvent: session context fields
+ *  - store attribution: pi-ingested rows carry source = "pi" (no opencode rows)
+ */
+
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { connect } from "./db.js";
+import {
+	buildIngestPayloadFromPiEvent,
+	buildPiFlushSignalFromEvent,
+	buildRawEventEnvelopeFromPiEvent,
+	MAPPABLE_PI_EVENTS,
+	mapPiEventPayload,
+} from "./pi-hooks.js";
+import { ingestRawEvents } from "./raw-event-ingest.js";
+import { MemoryStore } from "./store.js";
+import { initTestSchema } from "./test-utils.js";
+
+function requireEnvelope(envelope: ReturnType<typeof buildRawEventEnvelopeFromPiEvent>) {
+	if (envelope === null) {
+		throw new Error("expected pi envelope");
+	}
+	return envelope;
+}
+
+// ---------------------------------------------------------------------------
+// mapPiEventPayload — event type mapping
+// ---------------------------------------------------------------------------
+
+describe("mapPiEventPayload", () => {
+	describe("session_start → session_start", () => {
+		it("maps session start with deterministic id", () => {
+			const event = mapPiEventPayload({
+				piEvent: "session_start",
+				sessionId: "pi-sess-1",
+				cwd: "/tmp/repo",
+				ts: "2026-06-01T12:00:00Z",
+			});
+
+			expect(event).not.toBeNull();
+			expect(event?.schema_version).toBe("1.0");
+			expect(event?.source).toBe("pi");
+			expect(event?.event_type).toBe("session_start");
+			expect(event?.session_id).toBe("pi-sess-1");
+			expect(event?.event_id).toBe("pi:pi-sess-1:session_start");
+			expect(event?.cwd).toBe("/tmp/repo");
+			expect(event?.ts).toBe("2026-06-01T12:00:00Z");
+		});
+
+		it("prefers entryId for the event id suffix when present", () => {
+			const event = mapPiEventPayload({
+				piEvent: "session_start",
+				sessionId: "pi-sess-1",
+				entryId: "entry-start-1",
+				ts: "2026-06-01T12:00:00Z",
+			});
+			expect(event?.event_id).toBe("pi:pi-sess-1:entry-start-1");
+		});
+	});
+
+	describe("session_shutdown → session_end", () => {
+		it("maps reason field", () => {
+			const event = mapPiEventPayload({
+				piEvent: "session_shutdown",
+				sessionId: "pi-sess-end",
+				reason: "user_exit",
+				ts: "2026-06-01T13:00:00Z",
+			});
+
+			expect(event).not.toBeNull();
+			expect(event?.event_type).toBe("session_end");
+			expect(event?.payload.reason).toBe("user_exit");
+			expect(event?.event_id).toBe("pi:pi-sess-end:session_end");
+			expect(event?.source).toBe("pi");
+		});
+	});
+
+	describe("message_end role=user → prompt", () => {
+		it("maps prompt text and meta", () => {
+			const event = mapPiEventPayload({
+				piEvent: "message_end",
+				sessionId: "pi-sess-msg",
+				entryId: "entry-u1",
+				role: "user",
+				text: "Run tests",
+				cwd: "/tmp/repo",
+				custom_field: "keep-me",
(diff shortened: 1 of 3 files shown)


packages/core/src/pi-hooks.ts
```diff
@@ -0,0 +1,450 @@
...
+		const reason = field(payload, "reason");
+		eventType = "session_end";
+		eventPayload = { reason: reason ?? null };
+		idPart = entryId || "session_end";
+		consumed.add("reason");
+	} else if (piEvent === "message_end") {
+		const role = coerceString(field(payload, "role")).toLowerCase();
+		const text = coerceString(field(payload, "text", "content", "prompt"));
+		if (!text) return null;
+		if (!entryId) return null;
```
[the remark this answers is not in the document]
ZeR020: Correct about Pi 0.41's native `message_end` shape — but this adapter intentionally never sees the raw pi event. The pi extension's `pi.on("message_end")` handler unwraps `event.message` (role/text via `extractMessageRole`/`extractMessageText`) and synthesizes a deterministic `entryId` (`stableMessageEntryId`, timestamp-discriminated) before POSTing the flat payload this mapper consumes (`packages/pi-extension/src/index.ts`, `payloads.ts` — landing in a later layer of this stack, see #1430). Keeping the unwrap on the producer side is what lets core stay pi-agnostic. Traced end-to-end: no caller passes the raw nested event. No change needed here.

packages/core/src/pi-hooks.ts
```diff
@@ -0,0 +1,450 @@
...
+	return false;
+}
+
+/**
+ * Deterministic event id: `pi:<sessionId>:<entryId|toolCallId|stable-suffix>`.
+ * Same logical event must yield the same id across HTTP/CLI retries.
+ */
+function buildPiEventId(sessionId: string, stablePart: string): string {
+	const part = stablePart.trim();
+	if (!part) throw new Error("pi event id stable part is required");
```
kunickiaj: Raw ingestion validates event_id against /^[A-Za-z0-9._:-]+$/ and a 128-character limit, but this concatenates unconstrained session and tool/message IDs. For example, sessionId="session/with/slash" and entryId="entry+1" produces an ID that ingestRawEvents rejects with RawEventIngestValidationError. Pi toolCallId is typed only as string, so provider-specific IDs can trigger this even if session IDs are UUIDs. Please hash the identity components into a fixed-length safe ID, as the Claude and Codex adapters do, and add charset/length regression coverage.

packages/core/src/pi-hooks.ts
```diff
@@ -0,0 +1,450 @@
...
+
+	if (piEvent === "session_start") {
+		eventType = "session_start";
+		eventPayload = {};
+		idPart = entryId || "session_start";
+	} else if (piEvent === "session_shutdown") {
+		const reason = field(payload, "reason");
+		eventType = "session_end";
+		eventPayload = { reason: reason ?? null };
+		idPart = entryId || "session_end";
```
kunickiaj: This fallback gives every shutdown for a session the same event ID. Pi can shut an extension down for reload and later shut the same session down for exit; the second event is then deduplicated, preserving the earlier reason/timestamp and losing the final lifecycle event. The continuation stack's payload builder also currently supplies the fixed entryId "session_end", so it hits the same collision. Please require or derive a retry-stable discriminator for each distinct shutdown occurrence and add a reload-then-exit regression test.

kunickiaj (changes requested):
Two ingestion-identity issues need correction before this adapter is safe to merge.

packages/core/src/pi-hooks.ts
ZeR020: Hashed identity into `pi_evt_` + sha256(sessionId|piEvent|idPart).hex.slice(0, 24), matching Claude/Codex. Added charset/length coverage, including `session/with/slash` + `entry+1`.

05d7b2f7

packages/core/src/pi-hooks.ts
ZeR020: Shutdown no longer uses a fixed `session_end` id. When `entryId` is missing or that generic value, the discriminator is `session_end:<reason>`. Reload then quit persist as two events; identical retries still dedupe.

05d7b2f7

ZeR020:
@kunickiaj requested identity fixes are in `05d7b2f7` (hashed event IDs + shutdown discriminator). CI is green. Ready for re-review.

ZeR020:
@kunickiaj ready for your review whenever you have a moment.

packages/core/src/pi-hooks.ts
```diff
@@ -0,0 +1,450 @@
...
+
+	if (piEvent === "session_start") {
+		eventType = "session_start";
+		eventPayload = {};
+		idPart = entryId || "session_start";
+	} else if (piEvent === "session_shutdown") {
+		const reason = field(payload, "reason");
+		eventType = "session_end";
+		eventPayload = { reason: reason ?? null };
+		idPart = entryId && entryId !== "session_end" ? entryId : `session_end:${coerceString(reason)}`;
```
kunickiaj: Two shutdowns with the same reason (for example, `/reload` twice) still produce the same ID, so the second is dropped. Please have the producer send a stable per-occurrence `entryId`—or use its stable payload timestamp as the fallback—and add reload → reload coverage.

packages/core/src/pi-hooks.ts
```diff
@@ -0,0 +1,450 @@
...
+		consumed.add("is_error");
+		consumed.add("error");
+		consumed.add("tool_error");
+	} else {
+		return null;
+	}
+
+	if (!idPart) return null;
+
+	const meta: Record<string, unknown> = {
```
kunickiaj, line 287: Please match the Claude and Codex adapters by exporting `PI_EVENT_ID_ALGO = "pi/1"` and storing it as `meta.event_id_algo`. This keeps stored IDs attributable if the algorithm changes.

packages/core/src/pi-hooks.test.ts
```diff
@@ -0,0 +1,824 @@
...
+			}),
+		);
+		expect(envelope.event_id).toMatch(/^[A-Za-z0-9._:-]+$/);
+		expect(envelope.event_id.length).toBeLessThanOrEqual(128);
+		const result = ingestRawEvents(store, envelope);
+		expect(result.inserted).toBe(1);
+		expect(result.skipped).toBe(0);
+	});
+
+	it("persists reload then exit as distinct session_end events", () => {
```
kunickiaj, line 804: This covers reload → quit, whose reasons already differ. Please add reload → reload: both occurrences should insert, while retrying either payload should skip.

kunickiaj (changes requested):
Thanks for addressing the original review. The hashed-ID fix looks good. One shutdown edge case still blocks approval; the other two comments are small identity-contract follow-ups that should fit naturally with that change.

packages/core/src/pi-hooks.test.ts
ZeR020: Added reload → reload coverage: both insert; retry of either payload skips.

packages/core/src/pi-hooks.ts
ZeR020: Generic `session_end` identity now includes the payload timestamp. Two `/reload` events with different `ts` insert as separate rows; retrying the same payload still skips. Unique `entryId` still wins when present.

Anything else needed on this review?

packages/core/src/pi-hooks.ts
ZeR020: Exported `PI_EVENT_ID_ALGO = "pi/1"` and stored it as `meta.event_id_algo`.

ZeR020:
@kunickiaj ready for your review whenever you have a moment. Second-round items are in (per-occurrence shutdown IDs, `pi/1` algo tag, reload→reload coverage).

GitHub Actions is green; Snyk is failing on this org's private-test quota.

ZeR020:
@kunickiaj whenever you have a moment, could you please take another look at this PR? I’ve addressed the review feedback from the previous round and pushed the fixes. Would appreciate a re-review when you get a chance.

kunickiaj (approved):

PR #1580, opened 2026-09-02
perf(sync): remove duplicate peer lookup and the scope-listing N+1

kunickiaj:
## Description

Follow-up to #1564 and #1565, which fixed `rows_read`. This one targets request volume, which is what `rows_written` is made of: every authenticated coordinator request writes two rows to `request_nonces` — one INSERT on the way in, one DELETE when it expires 600s later — so a redundant request costs writes no matter how little it reads.

Both changes are worth making on their own terms; neither is only a quota play.

**1. Duplicate `/v1/peers` per daemon tick.** `refreshCoordinatorPresenceForDaemon` called `refreshAuthorizedCoordinatorPeerTrust`, which fetches peers and stores their addresses — then discarded the peers it returned. Ten lines later `fetchCoordinatorStalePeers` re-read the config file, fetched the same `/v1/peers`, and repeated the same `refreshStoredCoordinatorPeerAddresses` write. Two identical round trips seconds apart, with the same side effect applied twice.

`refreshCoordinatorForDaemon` now returns the snapshot and the tick passes it through, so one tick makes one lookup. `fetchCoordinatorStalePeers` keeps its fetching path for callers without a snapshot — including the tick itself when the coordinator refresh threw, so a failed refresh still degrades to the old behaviour rather than silently skipping the preflight. The stale-set derivation moved into the exported `coordinatorStalePeerKeys` so both paths share one implementation. `refreshCoordinatorPresenceForDaemon` remains as a boolean-returning wrapper.

The snapshot is now taken before the `onAfterCoordinatorRefresh` callback rather than after it. For a best-effort offline preflight a few-seconds-old view is well inside tolerance.

**2. `GET /v1/scopes` N+1.** `requesterAuthorizedForScope` ran `listScopeMemberships(scope_id)` once per scope — reading every member of every scope to answer a question about one device. The endpoint now makes a single `listDeviceScopeMemberships` call, served by the already-present `idx_coordinator_scope_memberships_device_status` index, and matches scopes against that map. `(scope_id, device_id)` is the primary key so the mapping is 1:1, and the `membership_epoch >= scope.membership_epoch` check is unchanged. The helper had no other caller — `/v1/scopes/:scope_id/members` does its own check over memberships it already needs — so it is removed rather than left dead.

`listDeviceScopeMemberships` is added to the store contract and implemented in both the D1 and better-sqlite stores.

## Type of Change

## Testing

`pnpm run tsc` and `pnpm run lint` clean. `pnpm run test` reports 5,736 passed — five more than before this change — with one unrelated environmental failure (`packages/mcp-server/src/http.test.ts` cannot bind `::1` in this container; it passes in CI). `pnpm --filter @codemem/cloudflare-coordinator-worker test:worker` passes, 6 tests.

New coverage:
- Shared store harness: one device's memberships across several scopes, isolation from other devices' grants, `includeRevoked` behaviour, and the empty-`deviceId` guard. Confirmed running against **both** store implementations.
- `fetchCoordinatorStalePeers` derives the same stale set from a supplied snapshot, with `globalThis.fetch` stubbed to throw so a second lookup would fail the test.
- `refreshCoordinatorForDaemon` returns the peer snapshot with exactly one `refreshAuthorizedCoordinatorPeerTrust` call, and reports `peers: null` when the coordinator is unconfigured.

## Checklist

- [x] Self-review completed

---

### DIFF
--- a/packages/core/src/better-sqlite-coordinator-store.ts
+++ b/packages/core/src/better-sqlite-coordinator-store.ts
@@ -2117,6 +2117,24 @@ export class BetterSqliteCoordinatorStore implements CoordinatorStore {
 			.map((row) => rowToRecord<CoordinatorScopeMembership>(row));
 	}
 
+	async listDeviceScopeMemberships(
+		deviceId: string,
+		includeRevoked = false,
+	): Promise<CoordinatorScopeMembership[]> {
+		const normalizedDeviceId = clean(deviceId);
+		if (!normalizedDeviceId) throw new Error("deviceId is required.");
+		// Served by idx_coordinator_scope_memberships_device_status.
+		const statusWhere = includeRevoked ? "" : "AND status = 'active'";
+		return this.db
+			.prepare(`SELECT scope_id, device_id, role, status, membership_epoch, coordinator_id, group_id,
+					manifest_issuer_device_id, manifest_hash, signed_manifest_json, updated_at
+				 FROM coordinator_scope_memberships
+				 WHERE device_id = ? ${statusWhere}
+				 ORDER BY scope_id ASC`)
+			.all(normalizedDeviceId)
+			.map((row) => rowToRecord<CoordinatorScopeMembership>(row));
+	}
+
 	async listScopeMembershipAuditEvents(
 		opts: CoordinatorListScopeMembershipAuditInput,
 	): Promise<CoordinatorScopeMembershipAuditEvent[]> {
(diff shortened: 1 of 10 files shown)


kunickiaj:
### Merge activity

* **Sep 2, 6:41 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/kunickiaj/codemem/1580).
* **Sep 2, 6:42 PM UTC**: [Graphite](https://app.graphite.com/github/pr/kunickiaj/codemem/1580) rebased this pull request as part of a merge.
* **Sep 2, 6:44 PM UTC**: [Graphite](https://app.graphite.com/github/pr/kunickiaj/codemem/1580) rebased this pull request as part of a merge.
* **Sep 2, 6:48 PM UTC**: @kunickiaj merged this pull request with [Graphite](https://app.graphite.com/github/pr/kunickiaj/codemem/1580).

PR #1603, opened 2026-09-04
fix: preserve DNS hostnames in device display names

### LINKED ISSUE #1602: fix: preserve DNS hostnames in device display names
The Devices page is displaying my machine name incorrectly.

Actual hostname:
`emm-five.home.morgowicz.com`

Displayed value:
`emm five.home.morgowicz.com`

The hyphen in the hostname is being converted to a space. DNS and fully qualified hostnames should preserve their hyphens and dots when used as automatic device names.

Human-entered device names can continue to use friendly separator normalization, but hostname-shaped values should remain recognizable as hostnames.

Please add a regression test covering a hostname such as `emm-five.home.morgowicz.com`, while preserving the existing behavior for genuine human-readable device names.

nmorgowicz:
## Description

Fixes #1602.

Automatic device-name normalization currently converts hyphens and underscores to spaces. This turns a hostname such as `emm-five.home.morgowicz.com` into `emm five.home.morgowicz.com`.

Hostname-shaped values now preserve their DNS separators, while ordinary human-readable device names keep the existing friendly normalization behavior. The shared UI machine-label classifier recognizes hostnames as machine-shaped values as well.

## Type Change

## Testing

- [ ] Relevant checks pass locally (the full suite is affected by an existing Node 26 and Vitest localStorage compatibility issue in unrelated UI tests)

Passed locally:

- `pnpm exec vitest run packages/core/src/project-invite-identity.test.ts packages/ui/src/tabs/feed/data/helpers.test.ts`
- `pnpm run tsc`
- `pnpm run lint`
- `pnpm build`

The full test suite reported 130 failures in unrelated UI and environment-sensitive tests because Vitest workers do not provide browser localStorage under the installed Node 26.7.0 runtime. The focused identity tests pass.

## Checklist

- [x] Code follows project style (pnpm run lint passes touched files)
- [x] Self-review completed
- [ ] Documentation updated (not needed for this bug fix)


packages/core/src/project-invite-identity.ts
```diff
@@ -37,6 +39,7 @@ export function isHumanPresentationName(value: unknown): boolean {
 		MACHINE_SLUG_NAME,
 		PENDING_PERSON_NAME,
 		HEX_HOSTNAME,
+		HOSTNAME_NAME,
```
kunickiaj: **P1:** `friendlyDeviceName()` now returns the FQDN, but `isHumanPresentationName()` rejects it. The coordinator therefore returns `device_display_name_invalid` during invite acceptance. Please keep generated hostnames valid through coordinator validation and add a round-trip regression test.

kunickiaj (changes requested):

packages/core/src/project-invite-identity.ts
nmorgowicz: Thanks, good catch. 

You’re right that the generated FQDN was being rejected during coordinator invite acceptance. I added device-specific validation that accepts valid DNS hostnames, updated both coordinator acceptance paths, and added a project-invite round-trip regression test. Human and recipient names still use isHumanPresentationName() and continue rejecting machine-shaped values. Pushed in 7667cb1.

If you have a better idea, please feel free to edit - i just wanted to make it easier on you. :)

packages/core/src/project-invite-identity.ts
```diff
@@ -59,11 +72,19 @@ export function friendlyDeviceName(input: {
...
-			.replace(/\.local$/iu, "")
-			.trim();
+		const candidateText = String(candidate ?? "").trim();
+		const raw = candidateText.replace(/\.local$/iu, "");
 		// Classify before separator normalization so UUIDs and machine slugs cannot
 		// be disguised as human labels by replacing hyphens or underscores.
+		if (HOSTNAME_NAME.test(candidateText)) {
+			try {
+				const normalized = normalizeDeviceNameHint(candidateText);
+				if (normalized) return normalized;
```
kunickiaj, line 81: **P1:** This returns before validating the stripped name. For example, `a1b2c3d4e5f6.local` becomes `a1b2c3d4e5f6`, which the coordinator rejects. Please validate `normalized` with `normalizeDeviceDisplayName()` before returning, and add a `.local` machine-ID regression test.

nmorgowicz: Thanks, I’ve addressed the original issue by allowing valid DNS hostnames through device-name validation, updating both coordinator invite-acceptance paths, and adding a project-invite round-trip regression test.

I also applied your follow-up by validating the name after stripping .local, with a regression test for machine IDs such as a1b2c3d4e5f6.local. This is pushed in c1266bb.

One clarification: should .local hostnames be preserved as first-class device names instead? For example, should a1b2c3d4e5f6.local remain intact rather than being stripped to a1b2c3d4e5f6? I apologize, the request wasn't 100% clear to me.

packages/core/src/project-invite-identity.ts
kunickiaj: **P1:** This generic predicate validates person and team names, so adding `HOSTNAME_NAME` now rejects dotted labels like `jane.doe` and `Acme.Inc`. Please remove it here and keep hostname allowance scoped to `normalizeDeviceDisplayName()`, which already checks it directly.

packages/core/src/coordinator-api.test.ts
```diff
@@ -1619,7 +1619,7 @@ describe("createCoordinatorApp dependency injection", () => {
 				fingerprint: fingerprintPublicKey(publicKey),
 				recipient_actor_id: "actor-brian",
 				recipient_display_name: "Brian",
-				device_display_name: "Brian's Mac",
+				device_display_name: "emm-five.home.morgowicz.com",
```
kunickiaj: **P2:** This test fails on `unexpected_project_invite_fields` before device-name validation, so the FQDN is never exercised. Please add positive join coverage that reaches both coordinator acceptance paths and preserves the hostname.

kunickiaj (changes requested):
The prior FQDN and `.local` findings are fixed. These are the remaining changes needed; normal CI should run before approval.

packages/core/src/project-invite-identity.ts
nmorgowicz: Thanks, addressed both remaining comments in 7863f13.

P1: Hostname handling is now scoped to normalizeDeviceDisplayName(). isHumanPresentationName() no longer rejects dotted human labels such as jane.doe or Acme.Inc. The .local path still validates the stripped value, so machine IDs such as a1b2c3d4e5f6.local are rejected safely.

packages/core/src/coordinator-api.test.ts
nmorgowicz: P2: Reworked the tests so the FQDN reaches both coordinator join acceptance paths instead of the legacy-invite field rejection. Both paths now positively verify preservation of emm-five.home.example.com.

Focused tests, TypeScript, and targeted Biome checks pass.

kunickiaj (approved):
Thanks for the contribution and for addressing the review feedback. The hostname preservation and coordinator validation paths look good.

PR #1637, opened 2026-09-06
fix(core): stop vec0 full scans starving the viewer under write lock

kunickiaj:
## Description
Root-causes the Viewer instability seen during v0.44.0-alpha.1 dogfooding. Two independent defects compounded:

**1. sqlite-vec `vec0` full scans under the write lock (the actual cause).** `vec0` indexes only `rowid = ?`; every metadata predicate (`model = ?`, `memory_id IN (...)`, `rowid > ?` range) is a full scan that materialises metadata for every row. Vector-migration cleanup ran N/250 such scans per pass **inside `IMMEDIATE` transactions**, then deferred on any concurrent write and retried every 5s. On the live 140k-row DB the cleanup had looped for 12 hours, pinning a core and holding the write lock so the Viewer got `SQLITE_BUSY` and died.

Fix: walk the `memory_vectors_rowids` shadow table (a real AUTOINCREMENT b-tree) for cursors, read metadata via indexed `rowid = ?` point lookups, hold the write lock only for deletes. Persist `stale_scan_exhausted` with a watermark taken from the scan's own last-examined rowid (not post-hoc `MAX(rowid)`, which can hide rows inserted during the prune). Rows above the watermark are re-checked; a legacy row invalidates the flag. Falls back to the old path when the shadow table is absent. Idle poll raised 5s → 60s.

**2. Sync-daemon crash on error-write.** When a tick failed and the error-state write *also* threw `SQLITE_BUSY`, the rejection escaped the timer as unhandled and killed the process, orphaning the maintenance worker. That orphan then blocked the next startup because worker cleanup ran *after* database init. Fix: error-state persistence is best-effort with console fallback, timer rejections are consumed, and trusted-worker termination now precedes `prepareViewerDatabase`.

## Live validation
Cleanup that had looped since 11:24 UTC completed in ~60s on the fixed build: `cleanup_pending: 0`, `stale_scan_exhausted: 1`, removed 102,721 stale rows. Worker dropped to 0% CPU; Viewer stayed HTTP 200 throughout.

## Type of Change
- [x] 🐛 Bug fix
- [ ] 🚀 Feature
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance
- [ ] 🧪 Testing

## Testing
- [x] 191 targeted tests pass (`vectors`, `vector-migration`, `vector-migration-queue-coverage`, `maintenance-worker-runtime`, `sync-daemon`, `serve`)
- [x] `pnpm run tsc`
- [x] Targeted Biome (pre-existing warnings only), `git diff --check`
- [x] New regressions: zero `SCAN memory_vectors` inside delete txns; exhausted flag short-circuits with zero vector statements; watermark race (verified to fail against `MAX(rowid)` code); `SQLITE_BUSY` on error-write does not crash and next tick recovers; `idleIntervalMs >= 60000`
- [ ] Full workspace suite not rerun for this PR

## Checklist
- [x] Self-review
- [x] Independent CodeReviewer approved (one blocker found and fixed: watermark race)

Tracking: codemem-1m48 (P0), codemem-29vi. AUTOINCREMENT rowid non-reuse verified empirically against vec0 0.1.9 including post-VACUUM.

### DIFF
--- a/packages/cli/src/commands/serve.ts
+++ b/packages/cli/src/commands/serve.ts
@@ -692,6 +692,12 @@ async function startForegroundViewer(invocation: ResolvedServeInvocation): Promi
 	if (invocation.dbPath) process.env.CODEMEM_DB = invocation.dbPath;
 	if (invocation.configPath) process.env.CODEMEM_CONFIG = invocation.configPath;
 	warnIfViewerExposed(invocation.host, invocation.port);
+	const dbPath = resolveDbPath(invocation.dbPath ?? undefined);
+	if (!(await terminateTrustedMaintenanceWorker(dbPath, { gracefulMs: 1000, forceMs: 5000 }))) {
+		p.log.warn(
+			"Existing maintenance worker is not trusted or did not stop; starting viewer anyway",
+		);
+	}
 	const preparedDb = prepareViewerDatabase(invocation.dbPath);
 
 	const observer = new ObserverClient();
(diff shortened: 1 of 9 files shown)


kunickiaj:
* **#1637** <a href="https://app.graphite.com/github/pr/kunickiaj/codemem/1637?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kunickiaj/codemem/1637?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

kunickiaj:
### Merge activity

* **Sep 7, 12:43 AM UTC**: @kunickiaj merged this pull request with [Graphite](https://app.graphite.com/github/pr/kunickiaj/codemem/1637).

PR #1678, opened 2026-09-10
feat(viewer-server): add /api/pi-hooks alias and memory tool routes

ZeR020:
@kunickiaj Third stacked PR for pi support (#1430). Layers 1–2 are merged (#1473, #1546), so this now sits directly on `main`.

## Why

Pi needs a dedicated ingest URL that lands on the 0.41 canonical seam, plus the missing HTTP memory-tool routes thin clients use.

## What

- `POST /api/pi-hooks`: bounded parse → `buildRawEventEnvelopeFromPiEvent` → force `source: "pi"` → awaited `ingestNormalizedEnvelope`. No `recordRawEvent` / `stripPrivateObj`.
- Memory-tool routes (`remember`, `timeline`, `expand`, `schema`, `search_index`, `explain`, `distill_candidates`).
- POSTs use the shared bounded JSON parser; kinds from core `REMEMBER_MEMORY_KINDS`; string `"false"` is false.
- Adjusted for #1546's hardened contract: pi event ids asserted in the `pi_evt_` hashed form, ingest awaited, and the remember-kinds set widened to `Set<string>` for filter checks.

Deliberate: MCP helper bodies stay copied in this layer. Extracting them into core is a follow-up (tracked in ZeR020/codemem#8), not this PR. Docs arrive on the final layer of this track.

## Test plan

- `vitest run packages/viewer-server/src/routes/memory-tools.test.ts` — 23/23
- `pnpm build`, `pnpm run lint`, `pnpm run tsc`, full `vitest run` — green

Related: #1430, #1429


ZeR020:
@kunickiaj The "TypeScript Test" build step failed twice with the same single error:

```
packages/mcp-server build: ../core/src/embeddings.ts(239,50): error TS2306:
File '…/packages/embeddings/dist/index.d.ts' is not a module.
```

**Diagnosis: pre-existing build-graph race, not branch code.** Every package build runs `tsc --build --force`, and `viewer-server` / `mcp-server` / `cli` all reference `core`, which references `embeddings` — so any two parallel lanes of `pnpm -r run build` re-emit `packages/embeddings/dist/*.d.ts` while another lane's tsc is reading it. The error is exactly a mid-rewrite read (file present, no exports).

Evidence it's not this branch: this exact commit builds green locally from a clean tree three times (including runs restricted to 4 CPUs to mimic the runner), main at the merge-base is green, and the diff only adds `viewer-server` route sources/tests — no build scripts, tsconfigs, or non-viewer-server files. The timing of the two parallel `tsc --build --force` passes (they start within ~30ms of each other) just landed unluckily here.

If it recurs, the root fix would be dropping `--force` from the package build scripts (or serializing the recursive build) so concurrent lanes don't re-emit each other's declarations — happy to send that as a separate PR if you want.

Could you rerun the failed job when you have a chance? Thanks!

packages/core/package.json
```diff
@@ -20,7 +20,7 @@
 	],
 	"scripts": {
 		"clean": "rm -rf dist tsconfig.tsbuildinfo",
-		"build": "pnpm exec vite build && pnpm exec tsc --build --force",
+		"build": "pnpm exec vite build && pnpm exec tsc --build",
```
kunickiaj: **P1:** Please do not remove `--force` without replacing its declaration-emission guarantee. I reproduced this by running the core build twice: the first build created `dist/index.d.ts`; the second exited 0 after Vite emptied `dist`, but `tsc --build` trusted the surviving `tsconfig.tsbuildinfo` and emitted no declarations. The same edit appears in embeddings, mcp-server, and viewer-server, so repeat/release builds can publish packages whose `types` entrypoints do not exist. Restore `--force`, or clear build info before declaration emission, and add a post-build assertion for each published `types` file.

packages/viewer-server/src/routes/raw-events.ts
```diff
@@ -336,5 +337,26 @@ export function rawEventsRoutes(getStore: StoreFactory, sweeper?: RawEventSweepe
...
 	});
 
+	// POST /api/pi-hooks — ingest pi extension events (compat alias)
+	app.post("/api/pi-hooks", async (c) => {
+		const result = await parseJsonObjectBody(c, MAX_RAW_EVENTS_BODY_BYTES);
+		if (result instanceof Response) return result;
+		const payload = result;
+
+		try {
+			const envelope = buildRawEventEnvelopeFromPiEvent(payload);
```
kunickiaj: **P1:** This route needs the same target guard and routing-field stripping as `/api/raw-events`, `/api/claude-hooks`, and `/api/codex-hooks`. Without `validateViewerTarget`, a Pi request intended for another DB is accepted into this store. Passing the raw payload here also lets `db_path` and `identity_target` fall into `meta.pi_fields`, persisting local path/identity routing data with the event. Please resolve `store`, validate with `{ requirePairedTargets: true }`, then call `buildRawEventEnvelopeFromPiEvent(untargetedPayload(payload))`; add mismatch and field-stripping tests.

packages/viewer-server/src/routes/memory-tools.ts
```diff
@@ -0,0 +1,773 @@
...
+ *
+ * Project scoping matches existing viewer routes (pack/memory/forget): only an
+ * explicit `project` arg applies. Unlike MCP tools, the viewer does not inject
+ * cwd/CODEMEM_PROJECT as an implicit default — the server process cwd is not a
+ * reliable client project signal. Callers (pi extension, CLI) pass project when
+ * they want scope.
+ */
+type FilterParse = { ok: true; filters: MemoryFilters | undefined } | { ok: false; error: string };
+
+function isFilterScalar(value: unknown): boolean {
```
kunickiaj: **P2:** This generic scalar check is too permissive for the declared MCP contract. It accepts numbers/booleans for string fields and mixed scalar arrays for string-array fields, so inputs such as `include_scope_ids: false` or `widen_shared_min_personal_results: "abc"` are silently coerced/ignored instead of returning 400. Please validate each field against its actual schema (or share the MCP schema) and add malformed scalar/array cases.

packages/viewer-server/src/routes/memory-tools.ts
```diff
@@ -0,0 +1,773 @@
...
+	const resolvedProject =
+		explicitProject !== undefined
+			? explicitProject || undefined
+			: cleanProject(defaultProject) || undefined;
+	if (resolvedProject) {
+		filters.project = resolvedProject;
+		hasAny = true;
+	}
+
+	for (const key of [
```
kunickiaj: **P2:** This filter allowlist and `SCHEMA_FILTER_NAMES` are hand-copied from the MCP schema. A future MCP filter can be omitted here silently; exclusion filters would then fail open and return broader results than requested. Please expose one shared key/schema definition for both surfaces, or at minimum add an exact parity test that fails whenever the lists diverge.

kunickiaj (changes requested):
Thanks for splitting this out. I found two blocking correctness/safety issues and two HTTP/MCP contract issues that should be addressed before merge. Details are inline.

packages/core/package.json
ZeR020: Fixed in 2872995b: each build now clears its own `tsconfig.tsbuildinfo` before `tsc --build`, so declarations always re-emit after vite empties dist, and asserts every published `types` entrypoint exists post-build (`dist/index.d.ts` + `dist/internal/cloudflare-coordinator.d.ts` for core; `dist/index.d.ts`/`dist/stdio.d.ts`/`dist/http.d.ts` for mcp-server). Verified with your double-build repro — the second run re-emits all declarations. I kept `--force` off because re-adding it reintroduces the parallel-lane collision that red CI'd this PR earlier (TS2306 mid-write on `packages/embeddings/dist/index.d.ts` when two `tsc --build --force` lanes re-emit the referenced project concurrently); clearing the package's own build info forces re-emission without foreign lanes rewriting referenced packages.

packages/viewer-server/src/routes/raw-events.ts
ZeR020: Fixed in d404fc55: the route now resolves the store, validates with `{ requirePairedTargets: true }`, and builds the envelope from `untargetedPayload(payload)` — same shape as `/api/raw-events`, `/api/claude-hooks`, and `/api/codex-hooks`. Added the `/api/pi-hooks` case to the targeted-ingest suites in index.test.ts (db/identity/contract mismatch → 409 before writes, partial targeting → 400, invalid types → 400, path-equivalent db accepted) plus a field-stripping test asserting `db_path`/`identity_target` never reach the persisted `payload_json`.

packages/viewer-server/src/routes/memory-tools.ts
ZeR020: Fixed in 4b0bc9ac: `buildFilters` now validates each field against its declared type from a shared catalog in `@codemem/core` (`packages/core/src/memory-filter-schema.ts`), so `include_scope_ids: false`, `widen_shared_min_personal_results: "abc"`, and mixed arrays like `include_visibility: ["private", 3]` return 400 instead of being coerced/ignored. Malformed scalar/array cases added to memory-tools.test.ts, including an acceptance case for the `boolean|string` union fields.

packages/viewer-server/src/routes/memory-tools.ts
ZeR020: Fixed in 4b0bc9ac: the filter key/type catalog now lives in one place, `@codemem/core`'s `memory-filter-schema.ts` (`MEMORY_FILTER_FIELD_TYPES` + `MEMORY_FILTER_NAMES`). The viewer derives both its request-validation allowlist and the `/api/memories/schema` filter list from that catalog directly, and the MCP zod shape is pinned to it by an exact parity test in `mcp-server/src/schemas.test.ts` (key equality plus per-type accept/reject, including the malformed scalar/array cases) — so a filter added on either surface without updating the catalog fails tests instead of failing open. I kept the MCP side as a literal zod shape held to the catalog by the test rather than deriving it programmatically: deriving would type the spread as `Record<string, ZodTypeAny>` and cost per-field inference in the tool args. If you'd prefer full derivation (or folding this into the shared tool implementations tracked in #8), happy to rework it.

kunickiaj (approved):
Re-reviewed at 4b0bc9ac6121c581140ff834f2f0d0ddacc22130. The four previously requested changes are resolved: repeat declaration builds clear stale TypeScript build metadata and assert published type entrypoints; Pi hook ingestion enforces paired targets and strips routing fields; HTTP memory filters validate values by field type; and the shared filter catalog is covered by MCP parity tests. Focused viewer/MCP tests pass (30 tests), the targeted ingestion mismatch cases pass (4 tests), and pnpm run tsc passes. No new blocking findings.

ZeR020:
@kunickiaj Thank you for the re-review. There is no remaining work from that review — this is ready to merge when you have a moment.

kunickiaj:
### Merge activity

* **Sep 12, 7:05 PM UTC**: [Graphite](https://app.graphite.com/github/pr/kunickiaj/codemem/1678) couldn't merge this PR because it failed for an unknown reason (Fast-forward merges are not supported for forked repositories. Please create a branch in the target repository in order to merge).
* **Sep 12, 7:07 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/kunickiaj/codemem/1678).
* **Sep 12, 7:07 PM UTC**: [Graphite](https://app.graphite.com/github/pr/kunickiaj/codemem/1678) couldn't merge this PR because it failed for an unknown reason (Fast-forward merges are not supported for forked repositories. Please create a branch in the target repository in order to merge).

PR #1744, opened 2026-09-16
feat(cli): add pi-hook-ingest/inject with shared spool

ZeR020:
@kunickiaj Fifth stacked PR for pi support (#1430). Layers 1–4 are merged (#1473, #1546, #1678, #1693), so this now sits directly on `main`.

## Why

Maintainer blocker: HTTP-success used to flush extraction before draining the spool, so compaction/shutdown could miss queued events.

## What

- `codemem pi-hook-ingest` / `pi-hook-inject` registered hidden in `command-tree.ts` (same as claude/codex hooks).
- Thin spool wrapper over shared `hook-ingest-spool.ts` (not a third lock/spool copy).
- Direct ingest: `buildRawEventEnvelopeFromPiEvent` + `ingestRawEvents`. No `recordRawEvent` / `stripPrivateObj`.
- HTTP-success + non-empty spool + boundary event: drain first, then flush.
- Inject fail-open (empty stdout on error).

Setup stays on layer 8. Boundary CLI timeout stays on layer 6.

## Test plan

- `vitest run packages/cli/src/commands/pi-hook-ingest.test.ts packages/cli/src/commands/pi-hook-ingest-durability.test.ts packages/cli/src/commands/pi-hook-inject.test.ts packages/cli/src/command-tree.test.ts` — 41/41
- `pnpm run tsc` — pass
- `pnpm run lint:delta -- --base origin/main --head HEAD` — pass

Related: #1430, #1429, #1693


kunickiaj (changes requested):
I found three blocking gaps in this slice. Focused validation otherwise passed (41/41 tests and `tsc --build`).

1. **[P1] Bind Pi HTTP ingest to the requested database and runtime identity** (`packages/cli/src/commands/pi-hook-ingest.ts:80-119, 379-404`). Unlike the Claude/Codex ingest paths, every HTTP attempt sends the raw Pi payload without `rawEventTarget(getDbPath())`. `/api/pi-hooks` accepts requests with no target fields, so `--db`/`--db-path` is silently ignored whenever any viewer answers on the configured port, and the event can be written under the viewer's different database/identity. Add the paired target fields to current and drained payloads and treat target-conflict 409s as a direct-only fallback, matching the existing adapters.

2. **[P1] Do not use the unscoped legacy GET pack endpoint as automatic fallback** (`packages/cli/src/commands/pi-hook-inject.ts:139-173, 175-197`). A local empty pack (not only a local error) falls back by default to `GET /api/pack`, which carries neither `db_path` nor `identity_target`; that legacy route performs no target validation. If a viewer for another database/identity is running, Pi can inject its memories into the current turn despite the explicit local `--db`. Use the profile-validated targeted POST transport used by Codex/Claude, or fail open with no injection when target identity cannot be proved.

3. **[P1] Recovered boundary events are deleted without replaying the boundary flush** (`packages/cli/src/commands/pi-hook-ingest.ts:259-293`). `deliverQueuedPiHook` reports success after HTTP/direct enqueue only; `drainPiHookSpool` then deletes the file. For `session_before_compact`, both enqueue paths deterministically skip because it is flush-only, so a boundary spooled while the DB/viewer was unavailable is later discarded without ever calling `boundaryFlush`. `session_shutdown` recovered through the direct path likewise gets stored without the synchronous flush this command promises. Drain handling must replay boundary semantics after durable delivery, with coverage where the original boundary flush fails and a later invocation recovers it.

ZeR020:
All three blocking gaps addressed in 98a09e4f (pushed to `stack/5-cli-ingest-inject`).

1. **HTTP ingest now bound to the requested target** — every HTTP attempt (current payload and drained spool payloads) is wrapped in `rawEventTarget(getDbPath())`, so `/api/pi-hooks` runs the same paired `db_path` + `identity_target` validation as the other adapters and `--db`/`--db-path` can no longer be silently ignored. `tryHttpIngest` surfaces 409 target conflicts (`viewer_db_mismatch` / `viewer_identity_mismatch` / `viewer_contract_unsupported`) as `targetMismatch`, which skips the locked HTTP retry and goes straight to the direct path — matching the Claude adapter.

2. **Inject fallback is now the profile-validated targeted transport** — the unscoped legacy `GET /api/pack` fallback is gone. `pi-hook-inject` fetches `GET /api/prompt-pack-profile` first, requires service/protocol-range compatibility plus exact `db_path` and canonical `identity_target` matches, and only then issues a targeted `POST /api/pack` carrying the same paired target fields. When a terminal disposition proves the endpoint serves a different database/identity, the command fails open with no injection (logged as `blocked=target_mismatch`) rather than risking another viewer's memories in the current turn.

3. **Boundary semantics replayed on drain** — `deliverQueuedPiHook` now runs the full boundary path after durable delivery of a spooled payload: `flushOnBoundaryIfRequested` performs the write-through direct ingest plus the `boundaryFlush` call for recovered `session_before_compact` / `session_shutdown` entries, so a boundary spooled while the viewer/DB was down is no longer deleted unflushed.

Coverage added: target fields present on HTTP payloads; 409 target-conflict skips the second HTTP attempt; boundary flush replayed when a failed original flush is recovered by a later drain (viewer-down-direct and viewer-up-HTTP paths); profile-mismatch injection block. 37/37 pi-hook tests pass; `tsc --build` and biome are clean on the touched files; full build passes.

kunickiaj (changes requested):
The two database/identity routing fixes look good. One blocker remains. Focused validation passes (46/46 tests and `tsc --build`).

**[P1] Do not delete a saved shutdown/compaction event until memory extraction succeeds** (`packages/cli/src/commands/pi-hook-ingest.ts:250-271, 299-312`).

When codemem cannot process a shutdown or compaction event, it saves that event to retry later. On the retry, the current code deletes the saved event as soon as it can deliver it—even if the memory-extraction flush still fails. That leaves nothing to retry, so memories from that session may never be extracted.

Keep the event in the spool when the flush fails, and remove it only after the flush succeeds. Add a test where the first retry still fails, confirms the event remains queued, and a later retry succeeds.

ZeR020:
Follow-up finding addressed in 615413de.

`flushOnBoundaryIfRequested` swallowed the replayed `boundaryFlush` failure, so `deliverQueuedPiHook` reported success and `drainPiHookSpool` deleted the entry — the extraction was lost one step later, exactly as described. Drain now uses a non-swallowing variant `flushOnBoundaryForRecoveredPayload`: a recovered boundary whose replayed flush throws is reported undelivered and stays spooled for a later invocation. Non-boundary payloads are unchanged (deleted once durably delivered — no flush semantics to lose).

Coverage added: spool a boundary with viewer+DB down → drain pass where delivery succeeds but the flush still fails (entry survives) → later drain where the flush succeeds (entry deleted). 38/38 pi-hook tests pass, `tsc --build` clean.

kunickiaj (changes requested):
The retry behavior is still broken in production, even though the new test passes. Focused validation passes (47/47 tests and `tsc --build`).

**[P1] Make the production flush report failure before deleting the saved event** (`packages/cli/src/commands/pi-hook-ingest.ts:163-223, 273-301, 329-343`).

When codemem retries a saved shutdown or compaction event, it now keeps the event only if `boundaryFlush` throws. The real `flushBoundaryRawEvents` function catches every observer, database, and extraction error and returns normally, so the retry code treats a failed extraction as success and deletes the saved event anyway. The new test uses a fake flush function that throws, which is why it does not reproduce production behavior.

Let the recovered-boundary path observe a real success/failure result from `flushBoundaryRawEvents` (while the normal hook path can remain fail-open). Delete the spool entry only after a successful flush, and cover the default production helper rather than only an injected throwing stub.

ZeR020:
@kunickiaj Addressed the production flush result.

`flushBoundaryRawEvents` now returns false when observer, store, or extraction fail (it still does not throw). The recovered-boundary drain deletes the spool only after that result is success. The live hook path stays fail-open.

The durability test no longer injects a throwing stub on the failing retry; it uses the default helper against a directory db path so store init fails without throwing.

Focused: 47/47 tests, `tsc --build` pass.

kunickiaj (approved):
The remaining durability blocker is fixed. The production flush now reports failures to the retry path, failed shutdown/compaction events stay queued, and the test covers failure followed by a successful retry. No remaining blocking findings.

Validated: 47 focused tests, TypeScript build, delta lint, and GitHub CI.

PR #1767, opened 2026-09-18
feat(pi-extension): add lifecycle client without native tools

ZeR020:
@kunickiaj Sixth stacked PR for pi support (#1430). Layers 1–5 are merged (#1473, #1546, #1678, #1693, #1744), so this now sits directly on `main`.

## Why

Maintainer blockers in this layer: project was cwd basename (not Git root); injection framing sniffed `## codemem memories` in untrusted pack text; boundary flush CLI used the short HTTP timeout.

## What

- New `@codemem/pi-extension` package **without** `tools.ts` (native tools are the next layer).
- `projectFromCwd` walks to `.git` / worktree gitdir; tests match core `resolveProject`.
- `fetchPackText` / file-context return `{ text, preformatted }`. No substring sniff.
- Boundary ingest CLI timeout 30s. File-context uses the 8s pack budget, not 3s.
- Workspace: `tsconfig` reference + lockfile importer.
- `@google/genai` allowBuilds: false (transitive dev-only dep of pi-coding-agent).

Version stays `0.40.1` until layer 8 (`0.41.0`).

## Test plan

- `vitest run packages/pi-extension` — 40/40
- `tsc -p packages/pi-extension --noEmit` — pass

Related: #1430, #1429, #1744


kunickiaj (changes requested):
I found two blocking data-isolation issues and one lifecycle leak. The focused tests (40/40), package typecheck, and package build pass, but the PR lint check is also currently failing.

1. **[P1] Prove the viewer database and runtime identity before sending or reading data** (`packages/pi-extension/src/client.ts:134-200, 241-301`). The direct HTTP paths send an unscoped payload to `/api/pi-hooks` and read from the legacy `GET /api/pack`. Neither request identifies the intended database or runtime identity. If another codemem viewer is listening on the configured port, Pi can write its session into that viewer's database and inject that viewer's memories into the current turn. This repeats the isolation bugs fixed in #1744. Use target-validated transport for both operations, or use the already-targeted CLI paths unless the viewer target can be proved. Add mismatch tests for ingest and pack retrieval.

2. **[P1] Honor `CODEMEM_VIEWER=0` for all HTTP operations** (`packages/pi-extension/src/client.ts:147-168, 242-247`; `packages/pi-extension/src/viewer.ts:136-169`). Disabling the viewer only makes `ensureViewerRunning` return false. Ingest still calls `checkIngestAvailable` and can POST events, while pack retrieval always calls `tryHttpPack`. The existing HTTP-pack test even uses `viewerEnabled: false` and expects the request to happen. When viewer use is disabled, skip every probe/HTTP request and go directly to the CLI fallback.

3. **[P2] Remove abort listeners after each CLI process finishes** (`packages/pi-extension/src/client.ts:372-411`). `execCodemem` adds an abort listener that is removed only if the signal aborts. Every successful CLI fallback leaves its listener and child-process closure attached to the session signal, so a long offline session accumulates listeners and eventually emits warnings. Remove the listener in the exec callback on both success and error, with a repeated-call test.

The GitHub `TypeScript Lint` check also fails because the delta ratchet reports an ambiguous `noExcessiveLinesPerFunction` function identity in `packages/pi-extension/src/index.test.ts`; that check must be made green before merge.

ZeR020 (commented):
f395f6f8 addresses the three review items.

1. HTTP ingest POSTs `/api/pi-hooks` with `db_path` + `identity_target`. Pack is GET `/api/prompt-pack-profile` (redirect: manual) then POST `/api/pack` with the same pair — no GET `/api/pack`. Profile/db/identity mismatch skips the pack POST and falls back to CLI. 409 `viewer_db_mismatch` / `viewer_identity_mismatch` / `viewer_contract_unsupported` is one HTTP POST then CLI, no retry.

2. `viewerEnabled: false` skips ensureViewer, checkIngestAvailable, tryHttpIngest, tryHttpPack, and proveAndPostPack. Zero fetch; CLI only.

3. `execCodemem` removes the abort listener in the execFile callback on success and error. Repeated successful execs then abort do not kill leftover children.

Identity is duplicated in `viewer.ts` (plugin keys) so the package still does not import `@codemem/core`.

kunickiaj (approved):
All requested fixes are complete. Pi now proves the viewer database and runtime identity before HTTP ingest or pack retrieval, `CODEMEM_VIEWER=0` skips all HTTP operations, completed CLI calls remove their abort listeners, and the lint-ratchet failure is fixed. No remaining blocking findings.

Validated: 48 focused tests, package typecheck, delta lint, package build, and GitHub CI.

PR #1778, opened 2026-09-19
feat(pi-extension): register native memory tools

ZeR020:
@kunickiaj Seventh stacked PR for pi support (#1430). Layers 1–6 are merged (#1473, #1546, #1678, #1693, #1744, #1767), so this now sits directly on `main`.

## Why

Maintainer blocker in this layer: file-context replaced string `tool_result` content (original read text disappeared).

## What

- Native `memory_*` tools (HTTP preferred, CLI fallback). No `pi-mcp-adapter` required.
- String `tool_result` content is normalized before file-context append; original text remains.
- Default tool project is the Git-root identity. `all_projects` still drops `project`.
- `pi.tools_mode=mcp-adapter` skips native registration. Duplicate native+MCP surface warns once.
- Tool HTTP skips fetch when `viewerEnabled` is false (CLI fallback).

Version stays `0.40.1` until layer 8 (`0.41.0`).

## Type of Change

- [x] 🚀 Feature (new functionality)

## Testing

- [x] Relevant checks pass locally (`pnpm run tsc`, `pnpm run lint`, targeted `vitest run packages/pi-extension`)

- `vitest run packages/pi-extension` — 60/60
- `tsc -p packages/pi-extension --noEmit` — pass
- `lint:delta --base origin/main` — pass

## Checklist

- [x] Self-review completed

Related: #1430, #1429, #1767


kunickiaj (changes requested):
I found three blocking gaps. The existing validation passes (60/60 focused tests, package typecheck, delta lint, and GitHub CI), but the new tool tests do not cover these transport and fallback transitions.

1. **[P1] Bind every native-tool HTTP operation to the requested database and runtime identity** (`packages/pi-extension/src/tools.ts:194-231, 281-775`). `httpJson` sends no `db_path` or `identity_target`, and these memory routes do not validate a target. `memory_pack` also reintroduces the unscoped legacy `GET /api/pack` path fixed in the previous layer. If another viewer answers on the configured port, Pi can read, create, or forget memories in that viewer's database. Add target validation at the operation endpoint (not only a separate probe), or use the targeted CLI path when the target cannot be proved. Cover mismatched viewers for both reads and writes.

2. **[P1] Do not blindly retry a possibly committed `memory_remember` through the CLI** (`packages/pi-extension/src/tools.ts:59-71, 543-586`). `httpJson` reports the same failure after a pre-send error and after a connection loss following a committed server write. `httpOrCli` always runs the CLI fallback, so the latter case creates the memory twice. Add an idempotency key understood by both transports, or distinguish safe pre-send failures from ambiguous outcomes and avoid replaying non-idempotent writes. Add a test where the HTTP write commits and the response is lost.

3. **[P1] Preserve the tool arguments promised by the schemas across every transport** (`packages/pi-extension/src/tools.ts:403-437, 440-616`). Several parameters are silently discarded: `kind` is ignored by pack/get/get-observations; forget ignores explicit `project` and `kind` filters entirely; and the CLI remember fallback always writes confidence `0.5` because it never forwards the requested confidence. The forget case can soft-delete an exact ID even when the caller supplied a nonmatching safety filter. Either implement each argument consistently in HTTP and CLI paths or remove unsupported arguments from that tool's schema, with execution tests for both transports.

ZeR020:
Addressed the three P1s from review 5256083010 (commit 6bf18b9a):

1. **Bind HTTP ops to db + identity** — every `httpJson` call now re-sends `db_path` + `identity_target` as query params, validated pre-handler by a new `viewerTargetQueryGuard` mounted on `/api/memories/*` and `/api/memory`. `memory_pack` no longer uses the unscoped `GET /api/pack`; it reuses `proveAndPostPack` (targeted POST). Unproved/mismatched viewers fall back to the targeted CLI path. `viewerEnabled:false` still skips every fetch. Mismatched-viewer tests cover reads (pack/get) and writes (remember/forget), including server-side 409 + no-write assertions.

2. **No blind remember replay** — `httpJson` now distinguishes safe pre-send failures (viewer disabled, connect refused, 4xx/409) from ambiguous post-send outcomes (timeout/reset/5xx). Only the former fall back to CLI; `memory_remember` surfaces an error instead of replaying after a possibly committed write. Test covers HTTP-commit-then-lost-response (exactly one write) and ECONNREFUSED (CLI fallback taken).

3. **Schema args on both transports** — `memory_remember` forwards `confidence` on CLI via a new `codemem memory remember --confidence` flag; get/get-observations forward `kind`+`project` on HTTP (expand honors them) and the CLI path rejects mismatches as `not_found`; `memory_forget` honors `kind`+`project` safety filters on HTTP (404, no delete) and CLI (show-then-skip); `memory_pack` keeps `project` (both transports) and drops the unsupported `kind` arg from its schema.

All gates green locally: full `lint`, `tsc` (all packages), `lint:delta`, and vitest for the changed packages (pi-extension 74, viewer-server 642+31 routes, cli memory 28).

kunickiaj (changes requested):
The duplicate-write and schema-argument findings are fixed, and the focused validation passes. One P1 identity-isolation blocker remains.

**[P1] Prove the viewer enforced the target before trusting the operation response** (`packages/pi-extension/src/tools.ts:237-292`). The client now appends `db_path` and `identity_target`, but older viewer versions ignore unknown query parameters and execute the request normally—as the comment at lines 253-255 notes. `ensureViewer()` only probes reachability, so a stale viewer left running after an upgrade is treated as healthy. Native tools can then read, create, or forget memories in that viewer's database while Pi accepts the successful response. Require a compatible target-aware server proof before issuing/trusting these operations (or use an endpoint/version contract that old viewers cannot accept), and add a test where an older viewer returns 2xx while ignoring the target.

Re-review validation: 107 focused tests, full TypeScript check, correction-delta lint, Pi package build, and GitHub CI all pass.

ZeR020:
Re #5257518478 (P1 identity isolation): fixed in 4801a632 — every native-tool HTTP op now requires a target-aware proof before the call is issued or trusted: the client GETs `/api/prompt-pack-profile` and matches the server-reported `db_path` + `identity_target` against the process target (`profileMatchesViewerTarget`, the same proof `/api/pack` already used). An older viewer has no profile route (or reports a different target), so its 2xx can no longer be accepted — the op falls back to the targeted CLI path instead, same as other safe HTTP failures. `ensureViewer()` still only probes reachability, but unproven reachability now means CLI, not trust.

Added the requested test (`tools.native-http.test.ts`, "falls back to CLI when an older viewer 2xx's while ignoring the target"): a legacy viewer serving `/api/memories/*` with 2xx and no profile route, asserting reads and writes resolve via CLI from the process DB. Focused tests (75), full tsc, lint:delta, and package build pass.

kunickiaj (approved):
The stale-viewer fix closes the remaining identity-isolation blocker. Native HTTP operations now require a matching target-aware profile before the operation, and the regression test confirms an older viewer that ignores targets and returns 2xx is not trusted. The earlier duplicate-write and argument-parity fixes also remain intact.

Validated at `6b616dfa59b09166247137d816fed6d3f3e3d2ba`: 75 Pi tests, package typecheck, correction-delta lint, package build, and green GitHub CI.

PR #1788, opened 2026-09-20
feat: add pi setup, UI, docs, and pi-extension release pin

### LINKED ISSUE #1429: feat: add pi agent support

ZeR020:
@kunickiaj Eighth and final stacked PR for pi support (#1430). Layers 1–7 are merged (#1473, #1546, #1678, #1693, #1744, #1767, #1778), so this now sits directly on `main`.

Closes #1429

## Why

Last stack layer: setup, UI, docs, and the 0.41.0 version pin (maintainer blocker #7).

## What

- Idempotent `codemem setup` pi target (`--pi-only`, `--pi-mcp`, `--pi-extension-path`). Never persists pi API keys.
- Official OpenAI/Anthropic base URLs are not written into observer config (keeps 0.41 Responses + default tier routing).
- UI copy includes pi in the shared-client list.
- Docs: `/api/pi-hooks` as compat alias over `ingestNormalizedEnvelope`; Git-root project; unledgered HTTP `GET /api/pack`; OAuth-only observer is explicit.
- `@codemem/pi-extension` is pinned at the repo-uniform (`0.46.0-alpha.2`) version and is in the release-version / publish set, matching the lockstep-versioning rule in #1429. It is also added to the release latest-guard list, since its first publish will be a prerelease.

## Type of Change

- [x] 🚀 Feature (new functionality)

## Testing

- [x] Relevant checks pass locally (`pnpm run lint`, `pnpm run tsc`, `pnpm build`, full `pnpm run test`)

- Full `pnpm run test` — 7693 passed / 5 failed; all 5 verified non-real: 3× mcp stdio-viewer port-38888 sandbox remap (passes on CI), 2× load flakes re-verified green in isolation (claude-hooks 68/68, raw-event-sweeper 88/88)
- `lint:delta --base origin/main` — pass (20 changed files)
- Targeted: `setup-pi.test.ts` 30/30, `node --test scripts/release-version.test.mjs` 11/11
- `node scripts/release-version.mjs check` — aligned at 0.46.0-alpha.2 (pi-extension included)

## Checklist

- [x] Self-review completed

Related: #1430 (stays open), #1767, #1778


ZeR020:
Correction to the Testing section: `release-version.mjs check` is not green on this branch — it reports drift (`pi_extension_package: 0.41.0` vs repo `0.46.0-alpha.2`). The 0.41.0 pin was authored when 0.41.0 was the next release; main has since moved to 0.46.0-alpha.2. Since `setVersion` now includes pi-extension, the next `release:version` run realigns all 14 tracked packages. If you'd rather the first published version of `@codemem/pi-extension` match the upcoming release (e.g. 0.46.0), happy to move the pin — the rest of the layer is unaffected either way.

ZeR020:
Follow-up to my comment above: moved the pin from 0.41.0 to the repo-uniform version (0.46.0-alpha.2) so release-version.mjs check passes aligned, lockstep versioning per #1429 holds, and codemem setup under core 0.46+ writes an npm settings pin that will exist at first publish. Also added @codemem/pi-extension to the release latest-guard list (first publish is a prerelease — exactly the case docs/versioning.md describes).

kunickiaj (changes requested):
I found two code blockers and one release gate. The existing checks pass, but these cases are outside the added coverage.

1. **[P1] Do not combine independently derived Pi observer fields with an explicit provider/model** (`packages/cli/src/commands/setup.ts:1037-1166`). `unsetObserverSlots` and `applyResolvedPiObserver` fill every missing slot independently. For example, if codemem already has `observer_provider: "anthropic"` but no model/base URL, setup can persist Pi's OpenAI model and custom OpenAI gateway under that Anthropic provider; environment-only provider/model overrides have the same problem. Core's `applyPiDerivedObserverFields` already preserves tuple compatibility: derive the full provider/model/endpoint set only when both identity fields are unset, or fill a missing model only when the explicit provider matches Pi. Make setup follow that rule and add partial file/env override tests.

2. **[P2] Document the targeted POST pack path that Pi actually uses** (`README.md:260`, `docs/architecture.md:382`, `docs/plugin-reference.md:329`). The merged extension calls `proveAndPostPack`, which proves `/api/prompt-pack-profile` and then sends targeted `POST /api/pack`; it no longer uses `GET /api/pack`. The new docs repeatedly describe the preferred Pi injection path as GET, which will mislead users debugging routing, target validation, or ledger behavior.

3. **[P1 release gate] Bootstrap the new npm package before relying on the tag workflow, and cover it in release tests** (`.github/workflows/release.yml:136-185`, `scripts/release-workflow.test.mjs:11-18`, `scripts/release-latest-guard.test.mjs:16-23`). `npm view @codemem/pi-extension` currently returns E404, while this repository's trusted-publisher procedure says a package must exist before trusted publishing can be configured. The release job clears `NODE_AUTH_TOKEN` and publishes through OIDC, so the next tag will fail at this new package unless the documented distinct bootstrap prerelease and trusted-publisher setup happen first. Please record/complete that gate before release. Also add Pi to both hard-coded published-package test lists; their current “every published package” assertions pass while omitting the package introduced here.

Validation at `ab5e8ae5d348c8f9bce4ccab6ff9d0d3111f16fa`: 37 targeted setup/UI tests, 30 release-script tests, release-version alignment, full TypeScript check, delta lint, Pi build and npm pack dry-run, UI build, and green GitHub CI.

ZeR020:
Review follow-up (re: pullrequestreview-5261132969) in a3ccfde7:

1. **Observer tuple rule** — `applyResolvedPiObserver` now mirrors core's `applyPiDerivedObserverFields`: the full provider/model/endpoint set is derived only when both identity fields are unset (env or file); otherwise a missing model is filled only when the explicit provider matches pi, and nothing is derived when a model is set without a provider. Added partial file/env override tests (`installPi — observer partial overrides` in setup-pi.test.ts: mismatched file provider, matching file provider, env-only provider, env-only model).

2. **Pack path docs** — README.md, docs/architecture.md, and docs/plugin-reference.md now describe the targeted `POST /api/pack` path (proves `GET /api/prompt-pack-profile` first) instead of `GET /api/pack`; the unledgered note is retained (pi sends no attempt metadata).

3. **Release gate** — added a "Bootstrap status" section to docs/trusted-publisher.md recording `@codemem/pi-extension` as pending bootstrap (E404 today; OIDC-only publish cannot create the package, so the bootstrap prerelease + trusted-publisher setup must precede the first `v*` tag that includes it — the section notes to update it once done). The bootstrap itself needs maintainer npm credentials, so that step stays with you. The package is now in both hard-coded published-package test lists, and the "verify runs after the last publish" assertion checks pi-extension as the actual last publish. Also fixed a stale "All six package versions" checklist line to seven.

Validated: 34 setup-pi tests (30 prior + 4 new), 101 setup-suite, 35 pi-hook, 30 release-script tests, biome (0 errors; pre-commit ratchet passed), tsc, pnpm build.

kunickiaj (changes requested):
The pack-path docs and release-test coverage are fixed, and the normal observer tuple cases now follow core. One edge of the original P1 remains.

**[P1] Do not persist a model selected through a temporary env provider beside a conflicting file provider** (`packages/cli/src/commands/setup.ts:1058-1072`). `explicitProvider = needs.envProvider || needs.fileProvider` gives the env override precedence. If the file contains `observer_provider: "anthropic"`, the environment temporarily sets `CODEMEM_OBSERVER_PROVIDER=openai`, and Pi resolves OpenAI, setup writes Pi's OpenAI model into the file while leaving its Anthropic provider unchanged. Removing the env override then exposes a persisted Anthropic/OpenAI tuple. Avoid persisting Pi-derived identity fields while provider/model env overrides are active, or evaluate writes against the persisted tuple separately. Add the conflicting file-provider + matching env-provider case so the saved file remains coherent after the env is removed.

The npm bootstrap remains a documented pre-release operational gate rather than a code correction. Re-review validation at `840d4b9857406f5e020bf6131284e4c86b40e61d`: 34 setup tests, 19 release tests, CLI typecheck, correction-delta lint, and green GitHub CI.

ZeR020:
The env-override edge of the P1 is fixed at 40f22ac0.

`applyResolvedPiObserver` now returns before any write while a provider or model env override is active (`needs.envProvider || needs.envModel`), and with no override active the explicit-provider comparison runs against the file provider alone — env precedence no longer drives the decision. Core can evaluate against the env-mashed config because it re-derives in memory per run; setup persists, so the write decision is keyed to the tuple that survives in the file once the override is removed.

Added the requested case beside the other partial-override tests: file pins `observer_provider: "anthropic"` while `CODEMEM_OBSERVER_PROVIDER=openai` matches pi's resolution — nothing pi-derived lands in the file, so it stays coherent after the env is dropped.

npm bootstrap stays as the documented pre-release operational gate. Validation at 40f22ac0: 35 setup tests, lint, tsc, and build all clean.

kunickiaj (approved):
The remaining observer-config blocker is fixed. Setup now skips all Pi-derived persistent observer identity writes while provider/model environment overrides are active, and the regression test covers a temporary OpenAI env override beside a persisted Anthropic provider. The prior pack-path documentation and release-test fixes remain correct.

Validated at `40f22ac0c797a5e0d18e3321dcbab8069fda7433`: 35 setup tests, CLI typecheck, correction-delta lint, and green GitHub CI. The documented npm bootstrap/trusted-publisher step remains a pre-release operational gate.

PR #1809, opened 2026-09-21
feat(pi): add session search and historical session import

ZeR020:
Search over stored pi conversation content (not just extracted memories), plus idempotent backfill of pre-install pi session history.

## What's here

- **Core search** (`packages/core/src/pi-session-search.ts`): lexical, most-recent-first scan over stored raw pi events (`source: "pi"`) — no new tables/indexes/migrations by design (v1 LIKE-scan with a documented FTS5 upgrade path at the scan site). Limit 1–20 (default 10), per-result snippet 100–4000 chars (default 1200), ~50KB output cap with explicit truncation markers and counts; explicit empty result on no match. Results carry `source`, session id, project, role, timestamp, snippet + truncation flags on every surface.
- **Historical import** (`packages/core/src/pi-sessions-import.ts` + `codemem pi-import-sessions`): walks `~/.pi/agent/sessions/**/*.jsonl` (honors `PI_CODING_AGENT_DIR`), ingests user/assistant text blocks through the standard raw-event pipeline with source `"pi"`. Deterministic ids mirror the live `stableMessageEntryId` formula so a live-captured session imports **zero rows** (unique-index collision) and re-imports are no-ops (per-file size/mtime skip via persisted state). `--extract` (off by default, cost note in help) drains imported sessions through the standard sweeper `flushRawEvents` — no bespoke pipeline; extracted memories carry pi attribution like live sessions.
- **Three surfaces, one contract**: `GET /api/pi/sessions/search` (viewer route), `memory_session_search` (15th native pi tool — HTTP-first with CLI fallback, absent in adapter mode, no promptSnippet by design), and `codemem pi-session-search` (`--json` parity tested against the route). `codemem setup` prints the explicit-backfill hint.
- **Docs**: README + plugin-reference + architecture updated in the same PR (supported matrix: pi-native 15 tools; **mcp-server intentionally stays at 14** until the shared-primitives follow-up).

## Design notes

- Event-id derivation is mirrored byte-identically between `packages/pi-extension/src/payloads.ts` and core (core can't depend on the extension package); both sides now carry cross-reference sync comments.
- Search scopes: user/assistant text only (thinking/tool-call/tool-result content skipped, per pi session format). Real session files persist numeric `message.timestamp`, which is the discriminator making live/import ids collide.
- Known interplay, deliberate: with a viewer running, the sweeper can pick up pending imported sessions like any other backlog even without `--extract` — import just never triggers the observer itself (flag documented at invocation).

## Gates (task 8.1)

- `pnpm run lint`: pass (ratchet green vs base; baseline warnings unchanged)
- `pnpm run tsc`: pass
- `pnpm run test`: 8293 passed / 14 failed / 4 skipped — the 14 failures are **byte-identical on pristine `origin/main`** (verified by running the full suite on detached base: memory-ownership-recovery ×9, mcp stdio-viewer ×3 [hardcoded port 38888 in local sandbox], claude-hooks ×1, project-scope ×1 suite interplay). This change adds ~48 passing tests and zero new failures.
- `pnpm build`: pass

Planned as OpenSpec change `add-pi-session-search` (proposal/spec/design/tasks); happy to walk through the artifacts on request.


packages/core/src/pi-sessions-import.ts
```diff
@@ -0,0 +1,390 @@
...
+			inserted: 0,
+			skipped: 0,
+			error: error instanceof Error ? error.message : String(error),
+		};
+	}
+}
+
+/** Default persisted size/mtime state location: a JSON file beside the database. */
+function defaultPiImportStatePath(dbPath: string): string {
+	return join(dirname(dbPath), "pi-import-sessions.json");
```
kunickiaj: [P1] Do not let this cache suppress imports into a different or recreated database. The path is shared by every DB filename in the directory and survives deletion/recreation of the original DB. I reproduced importing one file into `a.sqlite`, then importing the same history into fresh `b.sqlite` beside it: the second run reported `filesUnchanged: 1`, inserted zero rows, and left the fresh DB empty. Bind skip state to the actual destination DB (and invalidate it when that DB is replaced), or keep the optimization inside the DB; deterministic event IDs already make safe reprocessing possible.

packages/core/src/pi-sessions-import.ts
```diff
@@ -0,0 +1,390 @@
...
+	if (role !== "user" && role !== "assistant") return null;
+	const text = extractPiMessageText(message);
+	if (!text) return null;
+	const rawTs = entry.timestamp;
+	const ts = typeof rawTs === "string" && rawTs.trim() ? rawTs.trim() : null;
+	const rawDiscriminator = message.timestamp;
+	const discriminator =
+		typeof rawDiscriminator === "number" && Number.isFinite(rawDiscriminator)
+			? rawDiscriminator
+			: null;
```
kunickiaj: [P2] Reconstruct the live `n:<sequence>` fallback when `message.timestamp` is absent. Returning null here hashes only session/role/text, while the live extension uses `n:1`, `n:2`, etc. I reproduced two timestamp-less user turns with identical text: import inserted one and skipped one, and neither imported ID can collide with its live-captured counterpart. Track the same per-session missing-timestamp sequence used by `messageDiscriminator`, and add duplicate-turn/live-dedupe coverage.

packages/core/src/pi-session-search.ts
```diff
@@ -0,0 +1,266 @@
...
+ * ingest triggers when measured slowness on real histories says so.
+ */
+function loadCandidateRows(db: Database, tokens: string[], sessionId: string): PiSessionEventRow[] {
+	const likeClauses = tokens.map(() => "payload_json LIKE ? ESCAPE '\\'");
+	const params: unknown[] = tokens.map(likePattern);
+	let sql = `
+		SELECT r.stream_id, r.ts_wall_ms, r.payload_json, s.project
+		FROM raw_events r
+		LEFT JOIN raw_event_sessions s ON s.source = r.source AND s.stream_id = r.stream_id
+		WHERE r.source = 'pi' AND r.event_type = 'pi.hook' AND (${likeClauses.join(" OR ")})
```
kunickiaj, line 158: [P2] Restrict matches to the extracted conversation text, not the whole serialized envelope. `payload_json LIKE` also sees cwd, project, session metadata, and adapter fields, and `collectMatches` never checks the tokens against `textEvent.text`. I reproduced searching `metadata needle` where those words existed only in the session cwd; the API returned `conversation contains only ordinary words` as a match. Keep this prefilter if useful, but require at least one effective token in the extracted text before returning/counting the row.

packages/core/src/pi-session-search.ts
```diff
@@ -0,0 +1,266 @@
...
+	`;
+	if (sessionId) {
+		sql += " AND r.stream_id = ?";
+		params.push(sessionId);
+	}
+	sql += " ORDER BY r.ts_wall_ms DESC, r.id DESC LIMIT ?";
```
kunickiaj, line 174: [P2] Apply the project filter before this bounded LIMIT. It is currently evaluated only in `collectMatches`, so 5,000 newer candidate rows from other projects can fill the scan window and hide valid rows from the requested project, yielding an empty/truncated response despite matches. Add the project predicate to the candidate query (using the same matching contract or an equivalent prefilter) so the bound applies within the requested scope.

kunickiaj (changes requested):
I found four blocking correctness issues. The new targeted suites pass (47 tests), but they do not cover fresh-database cache reuse, timestamp-less duplicate turns, envelope-only query matches, or project filtering beyond the scan window.

packages/core/src/pi-sessions-import.ts
ZeR020: Fixed by keeping the optimization inside the database: processed-file size/mtime state now lives in a `pi_import_state` table in the destination DB, so skip state can never cross databases or survive DB deletion/recreation — a fresh or recreated database simply re-imports, deduped by deterministic event ids as before. The JSON state file and the `statePath` option are gone. Covered in `pi-sessions-import.test.ts`: import into `a.sqlite`, then the same history into fresh `b.sqlite` beside it (full re-import), then again after deleting/recreating `b.sqlite` (full re-import).

packages/core/src/pi-session-search.ts
ZeR020: Fixed: `collectMatches` now requires at least one effective token in `textEvent.text` (case-insensitive) before returning/counting a row; the `payload_json LIKE` clause is kept as the cheap prefilter. New test seeds a session whose cwd and project contain the query words while the conversation text does not, asserts an empty result, and keeps a positive control matching on the text itself.

packages/core/src/pi-session-search.ts
ZeR020: Fixed: the project predicate now runs inside the candidate query via `projectColumnClause("s.project", …)` — the shared SQL equivalent of the filter already used by memory search, which over-approximates `projectMatchesFilter` (the exact JS check still runs per candidate row) — so the LIMIT bounds only in-scope rows. New test fills the scan window with 1,005 newer matching rows from another project and asserts the target project's older match is still returned.

packages/core/src/pi-sessions-import.ts
ZeR020: Fixed: `coercePiSessionMessage` now reconstructs the live `n:<seq>` fallback, incrementing a per-session counter exactly like the extension's `messageDiscriminator` (counted only for gated user/assistant-with-text messages, in file order), so timestamp-less ids collide with their live-captured counterparts instead of each other. New coverage parses two identical timestamp-less turns to `n:1`/`n:2`, imports both into one DB, and dedupes against a live-captured `n:1` event.

ZeR020:
When you have a moment please check this out

ZeR020:
@kunickiaj When you get a chance, could you take a look at this?

